PR #9, opened 2021-03-16
feat(virtualoperator): Add go client

goock:
fix https://github.com/coreweave/virtual-server-operator/issues/7


virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+	// prepare config for kubevirt client, you need to set env variable, KUBECONFIG=<path-to-kubeconfig>/.kubeconfig
+	kubevirtClientConfig := kubecli.DefaultClientConfig(&pflag.FlagSet{})
+
+	// get the kubevirt client, using which kubevirt resources can be managed.
+	kubevirtClient, err := kubecli.GetKubevirtClientFromClientConfig(kubevirtClientConfig)
+	if err != nil {
+		log.Fatalf("Cannot obtain KubeVirt client: %v\n", err)
+	}
+
+	// Create a new VirtualServer with the name "my-virtual-server" to be deployed in the "default" namespace
```
YitzyD, line 75: In the comment: not the "default" namespace in all instances

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+	})
+	// Configure the root filesystem of the VirtualServer to clone a preexisting PVC namedubuntu1804-docker-master-20210210-ord1
+	// sourced in the vd-images namespace
+	err = virtualServer.ConfigureStorageRootWithPVCSource(vsv1alpha.VirtualServerStorageRootPVCSource{
+		Size:             "40Gi",
+		PVCName:          "ubuntu1804-docker-master-20210210-ord1",
+		PVCNamespace:     "vd-images",
+		StorageClassName: "ceph-ssd-2-replica",
+		VolumeMode:       corev1.PersistentVolumeBlock,
+		AccessMode:       corev1.ReadWriteOnce,
```
YitzyD, line 103: VolumeMode Block and AccessMode ReadWriteOnce are defaults. These are therefore optional and unnecessary in this instance. We could add a comment to note that or just leave it as is. Up to you

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+	})
+	if err != nil {
+		log.Fatalf("Cound not configure root filesystem\n")
+	}
+	// Expose tcp ports 22 and 443 on the VirtualServer
+	virtualServer.ExposeTCPPorts([]int32{22, 443})
+	// Expose a single udp port 4172 on the VirtualServer
+	virtualServer.ExposeUDPPort(4172)
+	// Set the VirtualServer to start as soon as it is created
```
YitzyD, line 112: We should add virtualServer.enablePublicIP(bool) and note that it will allow for the services to be public/private

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+	virtualServer.InitializeRunning(true)
+
+	// Delete virtual server if already exists
+	err = c.Delete(context.Background(), virtualServer)
+	if err != nil {
+		if apierrors.IsNotFound(err) == true {
+			fmt.Printf("VirtualServer %s in namespace %s already deleted", name, namespace)
+		} else {
+			log.Fatalf("Failed to create VirtualServer\nReason: %s", err.Error())
```
YitzyD, line 121: Failed to _delete_

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Create new virtual server
+	err = c.Create(context.Background(), virtualServer)
+	if err != nil {
+		log.Fatalf("Failed to create VirtualServer\nReason: %s", err.Error())
+	}
+
+	// Wait until virtual server is ready
+	fmt.Printf("VirtualServer status: %s", Ready(namespace, name, c))
```
YitzyD, line 132: I'm not sure this will work as expected. If the VS has yet to be created yet (for whatever reason) or has not been reconciled (so that the conditions are not yet present) this will return "return" or throw a fatal log and kill the process. When I think of Ready, I think of a function that blocks until it's ready, or takes a callback.

YitzyD (changes requested):

virtual-server/examples/go/main.go
YitzyD: Optionally, it could also block/callback for a specific VirtualServerConditionType with a status or a VirtualServerConditionReason.

virtual-server/examples/go/main.go
goock: as we talked about it already if the user does not set env var NAMESPACE, `default` namespace will be used.

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"kubevirt.io/client-go/kubecli"
+)
+
+// Ready waits until virtual server reach expected status
+func Ready(namespace, name string, c client.Client) string {
+	for {
```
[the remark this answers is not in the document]
goock: this one is still pending

virtual-server/examples/go/main.go
goock: Done

virtual-server/examples/go/main.go
goock: If they are default and I don't think we can change those 2 parameters for VM, I just removed them to not confuse users.

virtual-server/examples/go/main.go
goock: It works. The `Ready()` function is blocked until the required condition is fulfilled.

virtual-server/examples/go/README.md
```diff
@@ -0,0 +1,7 @@
+# Run examples
+
+The first run takes more time until all necessary packages are downloaded.
+
+```
+KUBECONFIG=/home/<user>/.kubeconfig NAMESPACE=<namespace> GO111MODULES=on go run main.go
```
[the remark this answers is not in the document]
goock: I added a very simple Makefile but I do not think it's a need to explain versions of the packages.  The puzzle is already solved in `go.mod` and `go.sum`.

goock:
@YitzyD This one is also ready to go, please review.

virtual-server/examples/go/main.go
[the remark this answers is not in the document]
goock: We agree on polling until the additional feature is implemented in vs controller.

salanki:
Can you please add some more details in the readme of what the example does?

goock:
> Can you please add some more details in the readme of what the example does?

I updated the `README.md`. The code has many comments and in most cases is self-explanatory.

YitzyD (changes requested):

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,160 @@
...
+// Ready waits until Virtual Server reach expected status
+func Ready(namespace, name string, c client.Client) string {
+	for {
+		vs := &vsv1alpha.VirtualServer{}
+		err := c.Get(context.Background(), client.ObjectKey{
+			Namespace: namespace,
+			Name:      name,
+		}, vs)
+		if err != nil {
+			log.Fatalf("cannot obtain virtual server status: %v\n", err)
```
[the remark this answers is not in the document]
goock: I use `continue` instead.

virtual-server/examples/go/main.go
```diff
@@ -0,0 +1,160 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"kubevirt.io/client-go/kubecli"
+)
+
+// Ready waits until Virtual Server reach expected status
+func Ready(namespace, name string, c client.Client) string {
```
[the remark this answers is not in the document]
goock: That's probably better solution to use consts here.

YitzyD (approved):

PR #20, opened 2021-07-14
VDI Example Terraform

jmarhee:
Provides an example for cloud users to:

- Deploy a virtual desktop through Terraform via module
- An example for consuming the Terraform module

Resolves https://github.com/coreweave/infra-pm/issues/84

### DIFF
--- a/virtual-server/examples/terraform/README.md
+++ b/virtual-server/examples/terraform/README.md
@@ -1,6 +1,6 @@
 # Deploying Virtual Servers to Kubernetes with Terraform
 
-This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` to [CoreWeave Cloud](coreweave.io).
+This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` to [CoreWeave Cloud](coreweave.com).
 
 ## Setup
 
--- a/virtual-server/examples/terraform/README.md
+++ b/virtual-server/examples/terraform/README.md
@@ -1,0 +1,22 @@
+# Deploying VirtualServers to Kubernetes with Terraform
+
+This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` (Virtual Desktops) to [CoreWeave Cloud](coreweave.io).
+
+## Setup
+
+This module requires your `user_namespace`, `kubeconfig_path`, your desired desktop `vdi_username` (and you can optionally supply `vdi_password` or set `vdi_password_generate` to `true`), `vdi_image` (defaults to Ubuntu 20.04), `vdi_gpu_enable` (and `vdi_gpu_count`), and your desired `vdi_name` to set your system hostname.
+
+## Installation
+
+Run:
+
+```bash
+terraform plan
+terraform apply -auto-approve
+```
+
+This module will output the network and credential information for the system, consumable by another module via the `network` and `password` attributes.
+
+## Examples
+
+In the `examples/` directory is a sample Terraform plan that demonstrates consuming the module to create two VDI instances.
--- a/virtual-server/examples/terraform/README.md
+++ b/virtual-server/examples/terraform/README.md
@@ -1,10 +1,10 @@
-# Deploying VirtualServers to Kubernetes with Terraform
+# Deploying Virtual Servers to Kubernetes with Terraform
 
-This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` (Virtual Desktops) to [CoreWeave Cloud](coreweave.io).
+This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` to [CoreWeave Cloud](coreweave.io).
 
 ## Setup
 
-This module requires your `user_namespace`, `kubeconfig_path`, your desired desktop `vdi_username` (and you can optionally supply `vdi_password` or set `vdi_password_generate` to `true`), `vdi_image` (defaults to Ubuntu 20.04), `vdi_gpu_enable` (and `vdi_gpu_count`), and your desired `vdi_name` to set your system hostname.
+This module requires your `user_namespace`, `kubeconfig_path`, your desired desktop `vs_username` (and you can optionally supply `vs_password` or set `vs_password_generate` to `true`), `vs_image` (defaults to Ubuntu 20.04), `vs_gpu_enable` (and `vs_gpu_count`), and your desired `vs_name` to set your system hostname.
 
 ## Installation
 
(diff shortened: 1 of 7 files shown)


jmarhee:
This is re: https://github.com/coreweave/GTM-Requests/issues/82

salanki:
Please [use a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) in the description to link the PR to the [tracking issue](https://github.com/coreweave/infra-pm/issues/84). That way the issue gets auto closed and you can also see the PR when looking at the issue. I updated the description with this.

virtual-server/examples/terraform/vdi.tf
```diff
@@ -0,0 +1,66 @@
...
+  manifest = {
+    "apiVersion" = "virtualservers.coreweave.com/v1alpha1"
+    "kind"       = "VirtualServer"
+    "metadata" = {
+      "name"      = var.vdi_name
+      "namespace" = var.user_namespace
+    }
+    "spec" = {
+      "initializeRunning" = true
+      "network" = {
```
salanki, line 13: Does it make sense to make this section configurable by the user as variables?

virtual-server/examples/terraform/vdi.tf
```diff
@@ -0,0 +1,66 @@
...
+          ]
+        }
+      }
+      "os" = {
+        "type" = var.vdi_os_type
+      }
+      "region" = var.vdi_region
+      "resources" = {
+        "cpu" = {
+          "count" = 3
```
salanki, line 37: This should be configurable just like GPU and RAM

virtual-server/examples/terraform/vdi.tf
```diff
@@ -0,0 +1,66 @@
...
+      "storage" = {
+        "root" = {
+          "size" = var.vdi_root_storage
+          "source" = {
+            "pvc" = {
+              "name"      = var.vdi_image
+              "namespace" = "vd-images"
+            }
+          }
+          "storageClassName" = "block-nvme-ord1"
```
salanki, line 54: The storage class name should be derived from the region. Ie, pseudocode: `block-nvme-${var.vdi_region}`.

salanki (changes requested):

salanki:
This is neat. I can see how this would evolve to Terraform modules provided by us.

virtual-server/examples/terraform/README.md
```diff
@@ -0,0 +1,22 @@
+# Deploying VirtualServers to Kubernetes with Terraform
+
+This [terraform](terraform.io) module uses the [kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs) to deploy `VirtualServers` (Virtual Desktops) to [CoreWeave Cloud](coreweave.io).
```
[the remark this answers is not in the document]
brianventuro: coreweave.com

virtual-server/examples/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
+variable "kubeconfig_path" {
+  description = "Path to kubeconfig"
+  default     = "~/.kube/config"
+}
+
+variable "user_namespace" {
+  description = "Namespace VDI will be installed to"
```
salanki, line 7: Change to Virtual Server

virtual-server/examples/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "vs_name" {
+  description = "VDI hostname"
+  default     = "MY-VDI"
```
salanki, line 12: VS

virtual-server/examples/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "vs_username" {
+  description = "VDI username"
```
salanki, line 16: VS

virtual-server/examples/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "vs_generate_password" {
+  type        = bool
+  default     = true
+  description = "Generate password"
+}
+
+variable "vs_memory" {
+  description = "VDI RAM"
```
salanki, line 32: VS

virtual-server/examples/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+  default     = "16Gi"
+}
+
+variable "vs_root_storage" {
+  description = "VDI root device storage (i.e 80Gi)"
```
salanki, line 37: VS

virtual-server/examples/terraform/vs.tf
```diff
@@ -0,0 +1,57 @@
...
+  manifest = {
+    "apiVersion" = "virtualservers.coreweave.com/v1alpha1"
+    "kind"       = "VirtualServer"
+    "metadata" = {
+      "name"      = var.vs_name
+      "namespace" = var.user_namespace
+    }
+    "spec" = {
+      "initializeRunning" = var.vs_running
+      "network" = {
```
salanki, line 13: Can we add in attachLoadBalancerIP here as well. It gets the Service IP directly attached to the VM (it becomes the Pod IP). It's a pretty common use case so good for completeness.

salanki (changes requested):
Some minor tweaks left for consistency on Virtual Server / VS. Once this is merged, please add a description and link this repo as a new entry under "Deployment Methods" in the documentation, with some CLI examples how to run it. You can access gitbook where we edit our docs via login.CoreWeave.com

salanki (approved):
@jmarhee: Can you give me some example output how this looks on the CLI?

jmarhee:
> @jmarhee: Can you give me some example output how this looks on the CLI?

The Kubernetes provider requires the attributes this module outputs to be marked sensitive for CRDs via this resource type, so I can't print them out to STDOUT (but are exported so another module can consume them) but what I can do is update the output to add a output message/command for the client to get their password from the state (and the IP from `kubectl`) -- would that be sufficient here (and update the documentation to reflect that as well)?

salanki:
So there is something in the Kubernetes provider limiting you to fetch the IP address from the VS status? Can you explain that more, seems unreasonable, no way to disable? If we need to instruct users, they should be doing `kubectl get virtualserver` not go after the service directly.

jmarhee:
Yes, this is being enforced by at the provider level, [as I understood it](https://www.terraform.io/docs/language/values/outputs.html#sensitive-suppressing-values-in-cli-output), for this resource type's exports, but can confirm. 

I can access it via the output CLI command since that reads from the state directly where it is in clear text; I can definitely look into un-enforcing this, because that would be much more useful for a case like this where we want users to have the information they need without going back into kubectl to get it (or I can just direct users back to the virtualserver resource in kubectl to get, both, the IP and credentials, etc. etc.).

jmarhee:
Actually, I have another method I can try to get the information through Terraform (from the provider, rather than try to export something from the resource), I'll try that first, and then see about unenforcing the sensitive value rules.

jmarhee:
The last push includes the above-mentioned workaround for retrieving the VS password, and extracting the Service IP from the Kubernetes provider, sidestepping the provider-enforced resource attributes. 

```
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

vs_network = "[address]"
vs_password = "[password]"
```

salanki:
Awesome. I'm gonna merge it, please write the docs in gitbook and close the tracking ticket once that is done!

PR #23, opened 2021-08-13
WIP: Updates kubectl examples to include: directAttach LB, shared PVC and …

jmarhee:
…claim example, Windows vs.

Signed-off-by: Joseph D. Marhee <jmarhee@coreweave.com>

resolves https://github.com/coreweave/infra-pm/issues/97

### DIFF
--- a/virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
+++ b/virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   accessModes:
   - ReadWriteMany
-  storageClassName: shared-nvme-ewr1
+  storageClassName: shared-nvme-ord1
   volumeMode: Filesystem
   resources:
     requests:
--- a/virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
+++ b/virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: your-pv-claim
+  namespace: my-namespace
 spec:
   accessModes:
   - ReadWriteMany
@@ -36,11 +37,12 @@ spec:
         pvc:
           namespace: vd-images
           name: ubuntu2004-docker-master-20210323-ord1
-    sharedFs:
-      enabled: true
-      name: sharedfs
-      claimName: your-pv-claim
-      mountPoint: /mnt/sharedfs
+    additionalDisks:
+      - name: sharedfs
+        spec:
+          persistentVolumeClaim:
+            claimName: your-pv-claim
+
 #  users:
 #    - username: SET YOUR USERNAME HERE
 #      password: SET YOUR PASSWORD HERE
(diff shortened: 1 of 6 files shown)


virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
```diff
@@ -0,0 +1,60 @@
...
+    memory: 16Gi
+  storage:
+    root:
+      size: 40Gi
+      storageClassName: block-nvme-ord1
+      source:
+        pvc:
+          namespace: vd-images
+          name: ubuntu2004-docker-master-20210323-ord1
+    sharedFs:
```
salanki, line 39: I don't think this is the right syntax for VS. did you test?

jmarhee: I hadn't meant to commit that file yet (because it hadn't worked-- I think I was basing this on an old template from kubeapps I had looked at for the TF example), but found what I needed, though this sample will need to be updated as well: https://github.com/coreweave/virtual-server-operator/blob/c30b0fcc3e30771edb2763ba9bfd71e0ecabec90/config/samples/_v1alpha1_virtualserver.yaml#L25-L33 because filesystems and additionalDisks (according to docs, and in testing) should be defined the same way

virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
```diff
@@ -0,0 +1,62 @@
...
+            claimName: your-pv-claim
+
+#  users:
+#    - username: SET YOUR USERNAME HERE
+#      password: SET YOUR PASSWORD HERE
+  network:
+    public: true
+    tcp:
+      ports:
+        - 22
```
salanki, line 53: Remove all ports except for 22 from all linux examples. The other ports are to support RDP.

virtual-server/examples/kubectl/virtual-server-with-pvc.yaml
```diff
@@ -0,0 +1,61 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: your-pv-claim
+  namespace: my-namespace
+spec:
+  accessModes:
+  - ReadWriteOnce
+  storageClassName: block-nvme-ord1
+  volumeMode: Filesystem
```
salanki, line 11: this should be `volumeMode: Block`

virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
```diff
@@ -0,0 +1,62 @@
...
+    memory: 16Gi
+  storage:
+    root:
+      size: 40Gi
+      storageClassName: block-nvme-ord1
+      source:
+        pvc:
+          namespace: vd-images
+          name: ubuntu2004-docker-master-20210323-ord1
+    additionalDisks:
```
salanki, line 40: `additionalDisks` is for block volumes. `filesystems` is for shared filesystems (via VirtioFS)

virtual-server/examples/kubectl/virtual-server-shared-pvc.yaml
goock: Removed

goock:
@salanki 
I change the manifest as you suggested.

Additionally I:
- changed the names, to be unique and users can run all examples at the same time,
- removed namespace, the manifests are ready to use, no changes are required,
- added users, these are examples so everybody wants to use it out of the box,
- added `README.md` with basic, yet important information
 
I tested all manifests.

salanki:
Is this ready?

goock:
> Is this ready?

@salanki yes, it's ready to go

PR #26, opened 2021-08-31
feat(virtualserver): Add examples

ChandonPierre:
This PR adds two new examples for kubectl: 

- [virtual-server-windows-cpu-only.yaml](virtual-server-windows-cpu-only.yaml)
-- Creates a CPU only Windows example, with reference to CPU type label selectors. Useful for servers that don't require GPU compute.
- [virtual-server-windows-internal-ip-only.yaml](virtual-server-windows-internal-ip-only.yaml)
-- Creates a `directAttachLoadBalancerIP: true` example with `public: false` - useful for servers like Domain Controllers


virtual-server/examples/kubectl/README.md
```diff
@@ -4,6 +4,10 @@ To run any of the examples issue: `kubectl apply -f <manifest_file_name.yaml>`
 
 - [virtual-server-direct-attach-lb.yaml](virtual-server-direct-attach-lb.yaml) shows how to directly attach the Load Balancer IP to a Virtual Server. This will give 
 
+- [virtual-server-windows-internal-ip-only.yaml](virtual-server-windows-internal-ip-only.yaml) creates a Windows Virtual Server with no public IP - useful for servers that will only be accessed in your namespace.
```
salanki, line 7: Clarify that this gets an internal STATIC ip (due to direct attach) and is useful for stuff like AD controllers.

ChandonPierre: Good call - updated in 7163997

ChandonPierre:
@salanki are we okay to merge this?

PR #29, opened 2021-09-03
feat(VirtualServer): Add cross-region cloning to pvc-clone.sh

ChandonPierre:
This PR adds the ability to clone a PVC between regions to `pvc-clone.sh`

### DIFF
--- a/virtual-server/pvc-clone.sh
+++ b/virtual-server/pvc-clone.sh
@@ -2,14 +2,14 @@
 # Clone the disk PVC for a given VM instance to a new PVC.
 set -e -o pipefail -u
 
-if [ $# -ne 2 ]; then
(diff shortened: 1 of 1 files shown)


ChandonPierre:
Yes, it is quite slow. The recommendation is still one base disk created per region, this just adds another option.

salanki:
Let's discuss how to improve this.

salanki:
I'm reopening because it would be great if we could solve this in a way that people can just clone via a DataVolume.

salanki:
Kyverno could possibly help us out here by adding an annotation on the copy-pod created by the CDI to run in the source region.

salanki:
@ChandonPierre: You just built this for linux, can we commit it to docs?

PR #36, opened 2021-09-17
fix(vs): Add missing param and update example

### LINKED ISSUE #35: Incorrect call to patch_namespaced_custom_object
https://github.com/coreweave/kubernetes-cloud/blob/03c2991d533a698dc296a827e340ace070afe46f/virtual-server/examples/python/vsclient.py#L56

This call needs an additional argument, `name`:
```
return self.api.patch_namespaced_custom_object(
            VSClient.GROUP, VSClient.VERSION, namespace, VSClient.PLURAL, name,manifest)
```

goock:
fix #35 

This PR adds a missing argument and adds an example with a manifest update.

### DIFF
--- a/virtual-server/examples/python/main.py
+++ b/virtual-server/examples/python/main.py
@@ -84,6 +84,18 @@ except ApiException as e:
 print(vsclient.create(my_virtualserver))
 print(f'VirtualServer status: {vsclient.ready(namespace, name)}')
 
+# Update the manifest and attach directly to Load Balancer
+my_virtualserver['spec']['network']['tcp']['ports'] = []
+my_virtualserver['spec']['network']['udp']['ports'] = []
+my_virtualserver['spec']['network']['directAttachLoadBalancerIP'] = True
+print(vsclient.update(my_virtualserver))
+
+# Restart the Virtual Machine Instance to apply changes.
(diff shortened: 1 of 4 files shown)


virtual-server/examples/python/main.py
```diff
@@ -84,6 +84,19 @@
...
 print(f'VirtualServer status: {vsclient.ready(namespace, name)}')
 
+# Update the manifest and attach directly to Load Balancer
+my_virtualserver['spec']['network']['tcp']['ports'] = []
+my_virtualserver['spec']['network']['udp']['ports'] = []
+my_virtualserver['spec']['network']['directAttachLoadBalancerIP'] = True
+print(vsclient.update(my_virtualserver))
+
+# Restart the Virtual Machine Instance to apply changes.
+print(vsclient.kubevirt_api.stop(namespace, name))
```
goock, line 94: The Kubevirt client we use does not have a restart command, I use stop/start, and additionally, I can wait until meeting the expected state.

YitzyD: We could add a restart function though, right? It would just be a 'restart' sent to the kubevirt_api

YitzyD (changes requested):

PR #38, opened 2021-10-07
terraform: update tf file defaults

cesarmesones:
- Remove gpu_enable option as by default this terraform code deploys a vm with a gpu.
- Update conditional for vs_attach_loadbalancer with ports, as specifying this to true requires empty tcp/udp ports.
- Casing on region string value.
- Remove deprecated provider kubernetes-alpha.


salanki:
@cesarmesones: Is this stale?

cesarmesones:
oh boy haha don't even know where I was going with this, but I would say yes..

PR #51, opened 2022-04-27
feat: Pin binary versions in Windows setup script

ChandonPierre:
Matches cluster version by default, adds switch to use latest binaries if desired


elantz1 (approved):
Sick! 8]

PR #83, opened 2022-07-21
Add bloom-175b inference example

Oogy:
Example to run [bigscience/bloom](https://huggingface.co/bigscience/bloom) as an inference service. 

**Note**: The full model has been successfully ran, but _not_ yet as an `InferenceService`. `Bloom-350m` has been successfully downloaded using a `Job`, and tested as an `InferenceService`, so the full model should work as well. The first test using the full model is in progress, but the download will take a while.

Fixes https://github.com/coreweave/infra-pm/issues/240


online-inference/bloom-175b/bloom-175b-inferenceservice.yaml
```diff
@@ -0,0 +1,59 @@
...
+                values:
+                  - A40
+        preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 10
+          preference:
+            matchExpressions:
+              - key: topology.kubernetes.io/region
+                operator: In
+                values:
+                  - LAS1
```
salanki, line 25: Do A40s in ORD1

online-inference/bloom-175b/bloom-175b-inferenceservice.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - "python3"
+          - "/workspace/bloom.py"
+        env:
+          - name: MODEL_NAME
+            value: 'bigscience/bloom'
+          - name: STORAGE_URI # Kserve mounts the PVC at /mnt/pvc/
+            value: pvc://bloom-175b-model-cache/
+          - name: TRANSFORMERS_CACHE
+            value: /mnt/pvc/
+          - name: MIN_LENGTH
```
salanki, line 39: Does this make sense? why have a min length larger than 1?

online-inference/bloom-175b/bloom-175b-inferenceservice.yaml
```diff
@@ -0,0 +1,59 @@
...
+            value: "50"
+          - name: MAX_LENGTH
+            value: "250"
```
salanki, line 42: Seems pretty short?

online-inference/bloom-175b/bloom-175b-inferenceservice.yaml
salanki, line 39: You can note with a comment that these are defaults and can be overridden in the request

online-inference/bloom-175b/model/requirements.txt
```diff
@@ -0,0 +1,3 @@
+accelerate
```
salanki, line 1: pin to a specific version

salanki (commented):
This is looking good so far, looking forward to the gitbook draft with step by step and `curl` instructions to run test inference!

PR #85, opened 2022-08-03
Fix unhandled condition in Python vsclient ready function

mysticmind:
In some cases, event object does not have `status`, handled this in Python `vsclient` ready function.


salanki:
Thank you for your contribution!

PR #88, opened 2022-08-19
Add stable-diffusion example

Oogy:


online-inference/stable-diffusion/03-inference-service.yaml
```diff
@@ -0,0 +1,61 @@
...
+        command:
+          - "python3"
+          - "/app/service.py"
+          - "--model-id=CompVis/stable-diffusion-v1-4"
+          - "--hf-home=/mnt/models/hub"
+          # The following arguments are defaults which can be changed here and per request
+          - "--precision=float16" # PyTorch dtype/model precision(one of float16 or float32)
+          - "--guidance-scale=7.0" # How well image generation adheres to the prompt. With very large numbers the images might look good, but will be less diverse. 
+          - "--num-inference-steps=50" # Results are better the more steps you use but will take longer.
+          - "--seed=42" # Every time you use the same seed you'll have the same image result. Use random numbers for different results with the same prompt.
```
salanki, line 35: Are these defaults or what's the purpose of these specified here?

salanki: Feels redundant since defaults are also specified in the API Code further down

online-inference/stable-diffusion/01-huggingface-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+data:
+  token: Q0hBTkdFTUU=
```
wbrown, line 3: Are we supposed to have a token here?

online-inference/stable-diffusion/02-model-download-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      - name: model-downloader
+        image: tweldoncw/model-downloader:10
+        imagePullPolicy: IfNotPresent
+        command: 
+          - "python3"
+          - "/app/download.py"
```
wbrown, line 14: Did you implement your own model downloader instead of extending the one we had? :)

online-inference/stable-diffusion/03-inference-service.yaml
wbrown, line 35: Why are we setting a fixed seed here?

online-inference/stable-diffusion/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+FROM nvidia/cuda:${CUDA_RELEASE} AS base
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt upgrade -y
+RUN apt update && apt install -y python3 python3-pip
+
+FROM base as isvc
+RUN mkdir -p /app
+ADD service/ /app/
+WORKDIR /app
+RUN pip3 install --upgrade pip
```
wbrown, line 11: Combine as many steps so that we don't have so many snapshots along the way.  pip3 should also be used with no cache.

online-inference/stable-diffusion/service/service.py
```diff
@@ -0,0 +1,124 @@
...
+logger = logging.getLogger(MODEL_NAME)
+logger.info(f'Model Name: {MODEL_NAME}')
+logger.info(f'Model ID: {options["MODEL_ID"]}')
+logger.info(f'Model Cache: {options["MODEL_CACHE"]}')
+
+parameters = {
+	'GUIDANCE_SCALE': float(os.getenv('CONDITION_SCALE', default=args.guidance_scale)),
+	'NUM_INFERENCE_STEPS': int(os.getenv('NUM_INFERENCE_STEPS', default=args.num_inference_steps)),
+	'SEED': int(os.getenv('SEED', default=args.seed)),
+    'WIDTH': int(os.getenv('WIDTH', default=args.width)),
```
wbrown, line 49: Run this through `black`.

wbrown (changes requested):

online-inference/stable-diffusion/01-huggingface-secret.yaml
salanki: Yeah good question what is this?

online-inference/stable-diffusion/03-inference-service.yaml
Oogy: If users want to change the defaults(and yes these values match the code defaults) they can see all options available and change them if desired in one place.

For other models we've included the defaults in the manifest as ENV vars, this is just a stylistic change to CLI arguments. 

Can change back to ENV vars, or remove as yes it is redundant(but IMO convenient).

online-inference/stable-diffusion/01-huggingface-secret.yaml
Oogy: The model has been released publicly, but you are still required to have a HuggingFace account to share your contact info and agree to the license before downloading.

Can be seen [here](https://huggingface.co/CompVis/stable-diffusion-v1-4) in a private/incognito window(assuming you have an HF account and are logged in).

The token is required to help enforce this. 

As for the specific value `Q0hBTkdFTUU=` this just base64 decodes to `CHANGEME` and will not actually work.

online-inference/stable-diffusion/02-model-download-job.yaml
Oogy: Yes, in the interest of time since there is no `huggingface_hub` lib for Go, and `diffusers` requires the `huggingface_hub` directory structure as described [here](https://huggingface.co/docs/huggingface_hub/v0.9.0.rc0/en/package_reference/file_download#caching). Also I'm more familiar with Python than I am Go :)

online-inference/stable-diffusion/03-inference-service.yaml
Oogy: All CLI args are here are to show what options are available. However this makes me realize, there is no way to use the default random seed. Will need to change the default for that in `service.py`.

online-inference/stable-diffusion/01-huggingface-secret.yaml
salanki: Aha, I assume that's in the docs how to encode it and apply it.

online-inference/stable-diffusion/03-inference-service.yaml
salanki: But defaults are here too right: https://github.com/coreweave/kubernetes-cloud/pull/88/files#diff-bf2cb0ebe5fe364e694c1d62e31788f44aebcb7db01d531fa306074e66e97346R17 feels duplicated?

online-inference/stable-diffusion/01-huggingface-secret.yaml
Oogy: Yes that is specified.

online-inference/stable-diffusion/03-inference-service.yaml
Oogy: yes it is duplicated. I personally like being able to see the options available from the manifest while I'm working on these because I'd prefer to change values from there when experimenting with different values. 

Some values can be changed per request, some cannot. I've rearranged args to properly indicate what can be changed and when. 

I can either remove these because they are admittedly redundant, or perhaps comment them out so the code is acting as the default but users can easily see what options are available just from the manifest if they want to change anything?

Oogy: Removed

Oogy: Now defaults to random seed with optional deterministic manual seed.

online-inference/stable-diffusion/README.md
```diff
@@ -0,0 +1,6 @@
+# Stable Diffusion
```
salanki, line 1: Needs to be updated with link to docs once published

online-inference/stable-diffusion/downloader/download.py
```diff
@@ -0,0 +1,13 @@
+# make sure you're logged in with `huggingface-cli login`
+import os
+import argparse
+import huggingface_hub as hf
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--model-id', default="CompVis/stable-diffusion-v1-3-diffusers")
```
salanki, line 7: Update to 1-4

salanki (approved):
Looking good, minor comments

salanki (approved):

online-inference/stable-diffusion/02-model-download-job.yaml
wbrown: @Oogy No problem, but we'll need to fix that. ;)

PR #91, opened 2022-09-16
feat (gpt-neox): Adding GPT-NeoX to Fastertransformer

rtalaricw:
This PR adds:

- [x] Download job and converter for GPT-NeoX weights
- [x] Inference Service with 2 RTXA_6000 GPUs 
- [x] Modifications to split GPTJ from GPT-NeoX


online-inference/fastertransformer/client/example.py
```diff
@@ -13,10 +13,16 @@
```
Oogy, line 13: Should encoding/decoding/tokenizing be the clients responsibility? Everything I've worked with so far that is handled server side, so all the clients need to do is send a curl request w/ some JSON which is simpler to demonstrate for examples.

rtalaricw: The goal was to isolate building the fastertransformer backend for Triton and not make it batteries included. The client examples are just for reference so that they can build out the client as per their requirements to go to production. Also there are two different models that we are encoding/decoding for so I thought it would be easier to see that hide away the implementation and leave them in the dark when they call a particular model on FT.

online-inference/fastertransformer/client/example.py
```diff
@@ -79,20 +91,21 @@ def prepare_tensor(client, name, input):
 
 def main(config, request):
     client_type = httpclient if config['protocol'] == 'http' else grpcclient
-    with client_type.InferenceServerClient(config['url'], verbose=config['verbose'], concurrency=10) as cl:
+    with client_type.InferenceServerClient(config['url'], verbose=config['verbose'], concurrency=1, connection_timeout=1000.0, network_timeout=1000.0,) as cl:
```
salanki, line 94: Why such long timeout? If that is in seconds that will make users wait a lot to get an error

online-inference/fastertransformer/download-weights-job-gpt-neox.yml
```diff
@@ -0,0 +1,287 @@
...
+          parameters {
+            key: "end_id"
+            value: {
+              string_value: "2"
+            }
+          }
+          parameters {
+            key: "data_type"
+            value: {
+              string_value: "fp32"
```
salanki, line 241: Seems as you are doing fp32 for NeoX too

online-inference/fastertransformer/client/example.py
rtalaricw: Will change to 100 seconds.

PR #94, opened 2022-09-25
BLOOM DeepSpeed Inference Example

Oogy:
Adds a deployment example for inference using the BLOOM 176B Model with Microsoft's DeepSpeed optimizations and pre-sharded weights. Requires 8x NVIDIA A100 80GB GPUs, optimally with NVLink.


online-inference/bloom-176b-deepspeed/files/isvc-patch.txt
```diff
@@ -0,0 +1,92 @@
+diff --git a/bloom-inference-server/models/model.py b/bloom-inference-server/models/model.py
```
salanki, line 1: what is this patch?

PR #108, opened 2022-11-28
Add a dataset downloader job to the Finetuning example

NavarrePratt:
This PR contains a bug fix in the argo workflow yaml and adds a kubernetes job to download a sample dataset to finetune on.

**Bug Fix**
The image that the argo workflow used for finetuning uses a `--fp16` flag to train in lower precision but the workflow was attempting to pass a value to the parameter: `--fp16 true`.
These changes will insert the `--fp16` flag to the container command based on whether or not the corresponding argo workflow parameter is set to `true`.

The first commit on this PR implements a fix by using multiple templates, one for fp16 and one without. The second commit, which I prefer, uses string templating to pass the entire command in as a command argument.

**Dataset Downloader**
Currently, the documentation assumes you already have a dataset ready to use when fine tuning. These changes (third commit) add a dataset downloader kubernetes job that will download book text files available on [SmashWords](https://www.smashwords.com/) which was used to create the original [BookCorpus](https://huggingface.co/datasets/bookcorpus) dataset and further variations (such as a part of [The Pile](https://pile.eleuther.ai/)).

The corresponding documentation changes can be found [here](https://app.gitbook.com/o/-M8EEYiNkaJsT8ISX0kX/s/-M83TghsCfsi8FCYs2DZ/~/changes/ORi5osiMLnnQZazjyzdG/compass/finetuning-machine-learning-models).


finetuner-workflow/dataset-downloader/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.19-alpine as source
```
wbrown, line 1: Can we switch the `dataset-downloader/Dockerfile` to using `ko`? You can see an example usage of this: here: https://github.com/wbrown/gpt_bpe/blob/main/.github/workflows/main.yml

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -158,50 +158,33 @@ spec:
...
-          - name: cache
-            value: "/{{workflow.parameters.pvc}}/cache/"
-          - name: torch_cache
-            value: "/{{workflow.parameters.pvc}}/torch/"
-          - name: no_resume
-            value: "{{workflow.parameters.no_resume}}"
-          - name: logs
-            value: "/{{workflow.parameters.pvc}}/{{workflow.parameters.logs}}"
+          - name: finetuner_params
+            # Pass the params in as one string so we can template in the conditional flags
```
wbrown, line 162: Nice workaround!

finetuner-workflow/dataset-downloader/Dockerfile
wbrown: We might want to get gitlab setup to build this/put this in ml-images. @todie, @salanki?

wbrown (changes requested):

finetuner-workflow/dataset-downloader/Dockerfile
NavarrePratt: Are you suggesting we get a build pipeline set up with this repo or moving the downloader to [the ml containers repo](https://github.com/coreweave/ml-containers) and building it from there?

PR #109, opened 2022-11-28
Add Stable Diffusion 2.0 support for SD Finetuning Workflow

harubaru:


wbrown (changes requested):
Bug to fix.

sd-finetuner-workflow/sd-finetuner/requirements.txt
```diff
@@ -1,4 +1,4 @@
-diffusers>=0.9.0
+git+https://github.com/huggingface/diffusers/
```
wbrown, line 1: This should really be pinned to a hash, instead of accepting whatever master is.

wbrown (changes requested):
Add a hash to the diffusers checkout. We don't want this breaking because of an update to master.

sd-finetuner-workflow/sd-finetuner/requirements.txt
wbrown: Also, you should be teliing it what the egg name will be.

i.e. the following example shows using a checkpoint + egg: `git+https://git@github.com/Dango233/diffusers.git@dango.dev.nov_update#egg=diffusers`

harubaru: Got it pointing to this commit which implements v-prediction: https://github.com/huggingface/diffusers/commit/6c56f05097f7d3c561f02dc1c27e3dd7e9f88ce1

PR #122, opened 2023-01-04
feat: Add Kubeflow training example

NavarrePratt:
This PR adds a Kubeflow distributed training example for a CNN on the MNIST dataset. It uses the PyTorch and MPI Operators.

The corresponding docs page is in review [here](https://app.gitbook.com/o/-M8EEYiNkaJsT8ISX0kX/s/-M83TghsCfsi8FCYs2DZ/~/changes/Jkl73wsADyJ7l0qZhbpz/compass/distributed-training-with-kubeflow).

### DIFF
--- a/kubeflow/mnist/k8s/mpi_job_mnist.yaml
+++ b/kubeflow/mnist/k8s/mpi_job_mnist.yaml
@@ -1,0 +1,70 @@
+apiVersion: "kubeflow.org/v2beta1"
+kind: "MPIJob"
+metadata:
+  name: "pytorch-horovod-mnist-mpi"
+spec:
+  slotsPerWorker: 4
+  runPolicy:
+    cleanPodPolicy: Running
+  mpiReplicaSpecs:
+    Launcher:
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        spec:
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_mpi_mnist:1
+              command:
+                - "mpirun"
+                - "-n"
+                - "4"  # Total processes = num workers * slots per workers
+                - "--allow-run-as-root"
+                - "python"
+                - "/opt/mnist/src/mnist.py"
+                - "--data-dir"
+                - "/mnt/pvc/dataset"
+                - "--model-dir"
+                - "/mnt/pvc/mpi/checkpoints"
+              resources:
+                requests:
+                  cpu: 2
+                  memory: 128Mi
+
+    Worker:
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        spec:
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_mpi_mnist:1
+              resources:
+                requests:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 4
+                limits:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 4
+              volumeMounts:
+                - name: kubeflow-mnist
+                  mountPath: /mnt/pvc
+          volumes:
+            - name: kubeflow-mnist
+              persistentVolumeClaim:
+                claimName: kubeflow-mnist
+          affinity:
+            nodeAffinity:
+              requiredDuringSchedulingIgnoredDuringExecution:
+                nodeSelectorTerms:
+                  - matchExpressions:
+                      - key: gpu.nvidia.com/model
+                        operator: In
+                        values:
+                          - RTX_A4000
+                      - key: failure-domain.beta.kubernetes.io/region
+                        operator: In
+                        values:
+                          - LAS1
--- a/kubeflow/mnist/k8s/pytorch_job_mnist.yaml
+++ b/kubeflow/mnist/k8s/pytorch_job_mnist.yaml
@@ -1,0 +1,109 @@
+apiVersion: "kubeflow.org/v1"
+kind: "PyTorchJob"
+metadata:
+  name: "pytorch-dist-mnist-nccl"
+spec:
+  pytorchReplicaSpecs:
+    Master:
+      replicas: 1
+      restartPolicy: OnFailure
+      template:
+        metadata:
+          annotations:
+            sidecar.istio.io/inject: "false"
+        spec:
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_dist_mnist:1
+              args:
+                - "--backend"
+                - "nccl"
+                - "--log-dir"
+                - "/mnt/pvc/pytorch/logs"
+                - "--data-dir"
+                - "/mnt/pvc/dataset"
+                - "--model-dir"
+                - "/mnt/pvc/pytorch/checkpoints"
+              resources:
+                requests:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 1
+                limits:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 1
+              volumeMounts:
+                - name: kubeflow-mnist
+                  mountPath: /mnt/pvc
+
+          volumes:
+            - name: kubeflow-mnist
+              persistentVolumeClaim:
+                claimName: kubeflow-mnist
+
+          affinity:
+            nodeAffinity:
+              requiredDuringSchedulingIgnoredDuringExecution:
+                nodeSelectorTerms:
+                  - matchExpressions:
+                      - key: gpu.nvidia.com/model
+                        operator: In
+                        values:
+                          - RTX_A4000
+                      - key: failure-domain.beta.kubernetes.io/region
+                        operator: In
+                        values:
+                          - LAS1
+
+    Worker:
+      replicas: 3
+      restartPolicy: OnFailure
+      template:
+        metadata:
+          annotations:
+            sidecar.istio.io/inject: "false"
+        spec:
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_dist_mnist:1
+              args:
+                - "--backend"
+                - "nccl"
+                - "--log-dir"
+                - "/mnt/pvc/pytorch/logs"
+                - "--data-dir"
+                - "/mnt/pvc/dataset"
+                - "--model-dir"
+                - "/mnt/pvc/pytorch/checkpoints"
+              resources:
+                requests:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 1
+                limits:
+                  cpu: 4
+                  memory: 16G
+                  nvidia.com/gpu: 1
+              volumeMounts:
+                - name: kubeflow-mnist
+                  mountPath: /mnt/pvc
+
+          volumes:
+            - name: kubeflow-mnist
+              persistentVolumeClaim:
+                claimName: kubeflow-mnist
+
+          affinity:
+            nodeAffinity:
+              requiredDuringSchedulingIgnoredDuringExecution:
+                nodeSelectorTerms:
+                  - matchExpressions:
+                      - key: gpu.nvidia.com/model
+                        operator: In
+                        values:
+                          - RTX_A4000
+                      - key: failure-domain.beta.kubernetes.io/region
+                        operator: In
+                        values:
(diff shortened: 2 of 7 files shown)


wbrown (approved):

kubeflow/mnist/k8s/mpi_job_mnist.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: "kubeflow.org/v2beta1"
```
salanki, line 1: don't use underscores in kubernetes manifest file names, and put the kind in the end,ie: `mnist-mpijob.yaml`

kubeflow/mnist/k8s/mpi_job_mnist.yaml
```diff
@@ -0,0 +1,70 @@
...
+                - "-n"
+                - "4"  # Total processes = num workers * slots per workers
+                - "--allow-run-as-root"
+                - "python"
+                - "/opt/mnist/src/mnist.py"
+                - "--data-dir"
+                - "/mnt/pvc/dataset"
+                - "--model-dir"
+                - "/mnt/pvc/mpi/checkpoints"
+              resources:
```
salanki, line 29: You want an affinity to schedule on a CPU. node in the same region as the worker (LAS1)

kubeflow/mnist/k8s/mpi_job_mnist.yaml
```diff
@@ -0,0 +1,70 @@
...
+                requests:
+                  cpu: 2
+                  memory: 128Mi
+
+    Worker:
+      replicas: 1
```
salanki, line 35: Make this 2 workers, since we are demoing distributed training.

kubeflow/mnist/k8s/pytorch_job_mnist.yaml
```diff
@@ -0,0 +1,109 @@
...
+              args:
+                - "--backend"
+                - "nccl"
+                - "--log-dir"
+                - "/mnt/pvc/pytorch/logs"
+                - "--data-dir"
+                - "/mnt/pvc/dataset"
+                - "--model-dir"
+                - "/mnt/pvc/pytorch/checkpoints"
+              resources:
```
salanki, line 27: Do launchers in PyTorch actually do compute? Ie it needs a GPU? Otherwise. Remove the GPU request/limit

kubeflow/mnist/k8s/pytorch_job_mnist.yaml
```diff
@@ -0,0 +1,109 @@
...
+                        operator: In
+                        values:
+                          - RTX_A4000
+                      - key: failure-domain.beta.kubernetes.io/region
+                        operator: In
+                        values:
+                          - LAS1
+
+    Worker:
+      replicas: 3
```
salanki, line 60: If my assumption is correct above and that the Master does work, it means there are total of 4 nodes working on this?

salanki (changes requested):
I would rename the directory either `training-operator` or `kubeflow-training-operator` as well.

salanki:
I would also make the example use full nodes, ie 8 GPUs (or do we build A4000 with 7), since that is what people should do for best performance.

salanki:
Screen Shot 2023-01-06 at 8 32 40 PM

As I read this, there is no data transfer fee for "Local", ie. if traffic is coming from the same region the "only" thing one has to pay is $50k/month.

kubeflow/kubeflow-training-operator/k8s/imagenet-pytorchjob.yaml
```diff
@@ -0,0 +1,157 @@
...
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_dist_resnet50:9
+              command:
+                - "torchrun"
+                - "--nproc_per_node=8"
+                - "--nnodes=$(WORLD_SIZE)"
+                - "--node_rank=$(RANK)"
+                - "--master_addr=$(MASTER_ADDR)"
+                - "--master_port=$(MASTER_PORT)"
```
NavarrePratt, line 24: @salanki - Yes, with `PyTorchJob` the master is doing compute as well. Not like the launcher in `mpijob`. In `PyTorchJob` the master and workers are pretty much the same except that `PyTorchJob` will set up the `MASTER_ADDR` and `MASTER_PORT` to point to the master.

kubeflow/kubeflow-training-operator/k8s/imagenet-mpijob.yaml
```diff
@@ -0,0 +1,100 @@
...
+          containers:
+            - name: pytorch
+              image: navarrepratt/pytorch_mpi_resnet50:1
+              command:
+                - "mpirun"
+                - "-np"
+                - "16"  # Total processes = num workers * slots per workers
+                - "-x"
+                - "WANDB_API_KEY=$(WANDB_API_KEY)"
+                - "--allow-run-as-root"
```
NavarrePratt, line 24: TLDR: I can't get good distributed performance with `MPIJob`.

To preface the below charts, the samples per second number come from `batch size / seconds spent on batch`. The numbers for 8/16 GPUs is an average across all the GPUs. Therefore if the 1 and 8/16 GPU samples per sec were the same, we'd have a perfect scaling factor.

These ran on A40 nodes with the following resource requests:
```yaml
resources:
  requests:
    cpu: 16
    memory: 128G
    nvidia.com/gpu: 8
  limits:
    cpu: 16
    memory: 196G
    nvidia.com/gpu: 8
```

This first chart shows the `MPIJob` performance on 1, 8, and 16 GPUs and clearly there's a big performance drop off. 
image

This next chart shows the `PyTorchJob` performance on 8 and 16 GPU. Much better performance and there's hardly a drop off for per GPU throughput when bumping up to 2 nodes.
image

I tried also setting the following `mpirun` flags according to [horovod's docs](https://horovod.readthedocs.io/en/stable/mpi_include.html#mpirun) but that didn't impact the performance. Anyone have a good guess to what's going on?
```
-bind-to none -mca pml ob1 -mca btl ^openib -x NCCL_SOCKET_IFNAME=eth0
```

kubeflow/kubeflow-training-operator/k8s/imagenet-mpijob.yaml
```diff
@@ -0,0 +1,100 @@
...
+              image: navarrepratt/pytorch_mpi_resnet50:1
+              resources:
+                requests:
+                  cpu: 16
+                  memory: 128G
+                  nvidia.com/gpu: 8
+                limits:
+                  cpu: 16
+                  memory: 196G
+                  nvidia.com/gpu: 8
```
NavarrePratt, line 75: I wasn't sure what the full cpu/mem values are on the A40 nodes. Should this be bumped up to whatever that is?

rtalaricw: Does it need 16 CPUs and 196Gi of memory to run the training?

kubeflow/kubeflow-training-operator/k8s/imagenet-pytorchjob.yaml
salanki: Ok, so that will give one extra node.

kubeflow/kubeflow-training-operator/k8s/imagenet-mpijob.yaml
salanki: What GPU types? You should have about 20-30% comms overhead going to multi GPU-multi node. I am surprised by the big drop-off you are having in the MPI Operator example, and the small one in the PyTorch operator. `NCCL_DEBUG=WARN` will show you how NCCL communicates. The only instance that would cause a big discrepancy is if it does very in-efficient intra-node comms in one instance but not the other. If you get scheduled on the same nodes in both cases that should not be a risk (you probably want to add affinities so you run on the same 2 nodes in each example to make sure it's not a node difference).

NavarrePratt: I was using A40's in LAS1 for both. I'll try running with `NCCL_DEBUG=WARN` and on the same two nodes for each type of job.

salanki: Add an affinity so you run on exactly the same nodes to weed out any inconsistencies.

NavarrePratt: Also for added context, because of how the PyTorch job has the master do compute, I go the single node (8 GPU) numbers for the second chart by running a normal k8s `Job` on a single 8 GPU container with the same `torchrun` command as you see in the yaml for the `PyTorchJob`.

For the MPIJob numbers I always ran it via `MPIJob` but I would change the number and resources of the worker containers.

NavarrePratt: Update:
With some changes described below, I was able to get the MPI Job's samples/sec to not drop off a cliff when going from 1 to 2 nodes. All of these numbers come from running on nodes `g08fb5e` and `g08fb06` with ~full node resources (90 cpu and 700G mem).

MPIJob:
image

PyTorchJob:
image

The change that increased the MPIJob's samples/sec and made it stable when going up to 2 nodes was increasing the resource request of the pods. Before, with 16 CPU and 128G mem, the GPU utilization was ~60-80% with 2 nodes and ~20-40% for 1 node. Bumping up the resources allowed the GPU util to get a stable >95% while training.

I initially decreased the resource requests down to 16 CPU, 64G mem, but needed to drop the data loader workers from 16 to 4 to get it to run with going OOM. That tanked a lot of metrics (samples/sec down from the >400 to 100-200). 

The previous PyTorchJob numbers were able to max out the GPU util which is why I think the 1/2 node numbers were already similar. Increasing the resources for the PyTorchJob just allowed it to go faster.

So the scaling factor going from 1 to 2 nodes is almost perfect. My guess as to why it's not as big of an overhead as the expected 20-30% is ResNet-50 is a much smaller model then we usually use with multi-node training. Lots less to communicate between nodes about. [Horovod's own benchmarks](https://horovod.readthedocs.io/en/stable/benchmarks_include.html#benchmarks) report a 90% scaling efficency up to 512 GPUs for ResNet101, double the size of resnet50, so I think our scaling efficiencies make sense. If we want to continue looking into it I can switch to a larger one of [torchvision's image classification models](https://pytorch.org/vision/stable/models.html#classification).

The main open question I still have is the samples/sec performance being almost double on the PyTorchJob vs MPIJob.

NavarrePratt: I was able to get the performance's of pytorch and horovod to line up after adding a `torch.cuda.synchronize()` before stopping the timer in the training loop. Noticed pytorch was doing that themselves in their [DDP benchmarks](https://github.com/pytorch/pytorch/blob/91920ee6da8b8566d0e4bdb4825bfcb4c4938ad5/benchmarks/distributed/ddp/benchmark.py#L78). I couldn't find any "hard" evidence as to why this is needed but I think it comes down to torch's DDP wrapping the model to handle the distribution and horovod wrapping the optimizer. Leads to slightly different race conditions when doing the quick timings for every training batch. 

Once I added that, all of the performance numbers line up. You can see my WandB charts [here](https://wandb.ai/npratt/resnet50-pytorch-vs-horovod) which has 1/2 nodes for torch/horovod.

rtalaricw: This slows down the PyTorch job by 2x though to align with MPI

NavarrePratt: The total runtime for pytorch and horovod were actually about the same the entire time (with and without `torch.cuda.synchronize()`, ~12 min for 3 epcohs with 16 GPUs.

The samples/sec numbers looks like it'll take twice as long but really it was a sneaky race condition that doesn't impact total runtime. However, I still included the synchronize in the `wandb` if statement since it's not needed if we didn't care about outputting accurate sampels/sec numbers.

rtalaricw (approved):
LGTM

wbrown (approved):
This looks good. Merging.

PR #128, opened 2023-01-19
Major LLM Finetuner Updates

wbrown:
Overview
---------
* Multi-GPU support.
* Support for recent models such as Pythia.
* Support tokenizer vocabularies other than `gpt` and `pile` provided by the model.
* Prompt testing while finetuning model.
* CUDA 11.8 and Torch 2.0

Multi-GPU Support
-------------------
* The `finetuner` now supports multiple GPUs on the same node, controlled by the `trainer_gpus` Argo workflow parameter.
* Also exposed are the `trainer_cores`, `trainer_ram` parameters as larger models and GPU counts will need commensurate increases in RAM and cores allocated.
* **TOFIX:** Currently works up to `7` GPUs. Adding an `8`th GPU exhausts a resource somewhere.
* **TODO:** Table of recommended CPU, RAM, and GPU configuration along with optimal batch sizes for each.

Support For Recent Models
---------------------------
* Huggingface's `transfomers` has been updated to the latest release version, `4.27.2` -- and the `finetuner` should support most any models that the `AutoModelForCausalLM` class supports.
* The `dataset_tokenizer` and `model_downloader` has been updated to work correctly with more models from the Huggingface Hub.
* The default model has been switched to `EleutherAI/pythia-2.8-deduped`.

Prompt Testing While Finetuning Model
---------------------------------------
* Often when finetuning a model, it is useful to be able to characterize the model during finetuning.
* A `prompt_file` parameter can be provided to read a file with each prompt on a line by itself.
* During finetuning, the `finetuner` will periodically run the prompts against the model (by default, on checkpoints), and send the output to `wandb` in addition to logging.
* The inference can be parameteried with `top_k`, `top_p`, `temperature`, and `repetition_penalty` Argo parameters.
* **TOFIX:** Inference is very slow when using DeepSpeed's ZeRO optimizer at `stage 3`, but `stage 2` yields incorrect results.

Library and Dependency Updates
----------------------------------
* PyTorch has been updated to the latest major release, `2.0`.
* CUDA has been updated to `11.8` to somewhat support the latest Hopper generation GPUs.
* The finetuner also can use a CUDA `12.0` based image, but this has not been as well tested.
* DeepSpeed has been updated to the latest version.

Bugfixes and Minor Additions
-----------------------------
* `TokenizedDataset` class now uses `mmap`.
* Initialize the random seed correctly and in the right place.
* We now report the estimated DeepSpeed memory usage on startup, for more easy tweaking.
* Added the ability to partition overlapping contexts with a `boundary_index` variable.
* `tokenizer` parameter actually works and falls back to the model default.
* Added a standalone `evaluator.py` script to run inference from `prompts_file`.
* Correctly work when there are multiple `finetuner` instances on the same node.
* Switched `finetuner` to use proper `logging`.
* Report more parameters on `finetuner` startup.
* Correctly calculate and report on memory utilization.
* Correct label masking.

Major Todo
-----------
* While the `finetuner` now works with multiple models, the inference endpoint code needs to be updated and cleaned up.


Rexwang8:
do you think for the inference settings, at the bottom, we should use 1 min pod and 6 cpu/64gb? I did this with mine and it helped but I'm not sure if that was just luck.

Rexwang8:
also is there a specific reason to downgrade transformers from 4.25 to 4.23?

Rexwang8:
also would it be possible to provide an example prompts.txt file

harubaru:
> also is there a specific reason to downgrade transformers from 4.25 to 4.23?, or just talking about the requirements.txt in general?

I believe it's because that the Tensorizer only supports transformers up to version 4.23

Rexwang8 (approved):

rtalaricw:
I am curious as to why 8 GPUs does not work? What are some thoughts or bugs that are encountered here?

finetuner-workflow/finetuner/ds_config.json
```diff
@@ -32,12 +32,14 @@
...
-        "stage3_gather_fp16_weights_on_model_save": true
+        "offload_optimizer": {
+            "device": "cpu"
+        },
+        "stage3_gather_16bit_weights_on_model_save": true
     },
     "gradient_accumulation_steps": "auto",
     "gradient_clipping": "auto",
-    "steps_per_print": 2000,
+    "steps_per_print": 1000000000000000,
```
rtalaricw, line 42: This looks odd to me. Is there an explanation why it went to this value from 2000?

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -374,14 +449,16 @@ spec:
       resources:
         requests:
           memory: 128Gi
-          cpu: "8"
+          cpu: 8
+          nvidia.com/gpu: 1
         limits:
-          memory: 192Gi
-          cpu: "8"
+          memory: 128Gi
```
rtalaricw, line 455: I have a feeling that setting this to 32/64 CPUs, with 256 GB memory might solve the 8 GPU issue.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -109,31 +142,39 @@ spec:
...
+      value: 192
     # Inference GPU - RTX A5000, 24gb VRAM
     - name: inference_gpu
       value: 'RTX_A5000'
     # Container images -- generally, don't alter this.
     - name: model_downloader_image
       value: 'ghcr.io/wbrown/gpt_bpe/model_downloader'
     - name: model_downloader_tag
-      value: 'e2ef65f'
+      value: '73cceb0'
```
rtalaricw, line 161: Can we set these tags dynamically for all images build by us?

rtalaricw (commented):
I think we should fix the 7->8 GPU issue since I am just curious what the root cause is. Other than that, few more comments...

finetuner-workflow/finetune-workflow.yaml
wbrown: No, it didn't. We did 704gb and 32 cores. It's not tied to RAM or CPU.

finetuner-workflow/finetune-workflow.yaml
wbrown: No, we don't want to do that. We want to control when we update our dependencies.

finetuner-workflow/finetuner/ds_config.json
wbrown: It doesn't allow disabling the logging, so we set it to an extremely high number.

microcosem (commented):
I will add this into the Finetuning Workflow docs and make a release note mention for it. Will create a new issue for that work.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -4,37 +4,53 @@
...
+import socket
+from contextlib import closing, contextmanager
+import logging
 import validators
+import deepspeed
+from deepspeed.runtime.zero.stage_1_and_2 import estimate_zero2_model_states_mem_needs_all_live;
+from deepspeed.runtime.zero.stage3 import estimate_zero3_model_states_mem_needs_all_live;
+
+
+def find_free_port():
```
harubaru, line 32: This function is unused, but remembering a previous conversation it's supposed to find a free port so multiple runs can be executed in the same environment. It might be good to test to see if it's needed or if deepspeed automatically finds a free port too, since I _think_ deepspeed supports that.

harubaru (changes requested):

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -306,35 +362,37 @@ spec:
     inputs:
       parameters:
```
Eta0, line 363: `workflow.parameters.retokenize` exists and is passed as `retokenize` to the `model-tokenizer` template earlier in the workflow, but the template itself neither defines it as a parameter nor uses it at the moment. In some earlier commits there was a commented out `-retokenize` flag passed along with the args to the container, but it seems to have disappeared.

finetuner-workflow/finetuner/requirements.txt
Eta0, line 1: A lot of the requirements in here are no longer used in the code. The following can be safely removed:
- `pandas`
  - Never used
- `tokenizers`
  - This is a already a dependency of `transformers` and we use it through that library instead; better to remove it to let the `transformers` dependency control the version restrictions
- `huggingface-hub`
  - Also a dependency of `transformers` and not used directly
- `validators`
  - Imported in `finetuner.py` but never used
- `requests`
  - Never used

finetuner-workflow/finetuner/evaluator.py
```diff
@@ -0,0 +1,213 @@
...
+except pynvml.nvml.NVMLError_LibraryNotFound:
+    pynvml = None
+
+device = torch.device("cpu")
+if torch.cuda.is_available():
+    device = torch.device("cuda")
+elif torch.backends.mps.is_available():
+    device = torch.device("mps")
+
+def get_gpu_ram() -> str:
```
Eta0, line 25: This should be updated to match the identically-purposed function in `finetuner.py` (which was tweaked a bit independently). It should be easiest to just separate it out from both of them into a `utils.py` file and import it, or to use the equivalent and even further updated [`tensorizer.utils.get_mem_usage` function](https://github.com/coreweave/tensorizer/blob/029982083a1278ed45f87cecbb675571d4343125/tensorizer/utils.py#L46-L109) when that dependency is integrated in.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -266,93 +645,102 @@ def __getitem__(self, idx) -> Tuple[torch.Tensor, torch.Tensor]:
...
 
-else:
-    try:
-        tokenizer = AutoTokenizer.from_pretrained(args.model,
-                                                  eos_token=args.eot,
-                                                  pad_token=args.pad,
-                                                  cache_dir=args.cache)
+try:
+    with no_init():
+        model_fp16_args = {"torch_dtype": torch.float16}
```
Eta0, line 674: `model_fp16_args` here looks like it was intended to only be passed conditionally in `FORCE_FP16` mode, but it's passed either way at the moment. Although, I actually [removed the downcast to fp16 here in an earlier commit](https://github.com/coreweave/kubernetes-cloud/commit/78e4a10243a483f5b162d4d573b268bfae2843dc) (when it was using `.half()`) even for the fp16 (actually mixed precision) mode after some testing showed that:
- It didn't make training or evaluation faster,
- It didn't decrease VRAM usage, and
- It's a completely invalid configuration with base the `transformers` library and only works because `deepspeed` has more lenient runtime checks

I think this was because DeepSpeed's mixed precision training casts most of the stored model weights back up to fp32 anyway, or something like that, so you lose precision with no particular benefit.
If casting it like this works substantially differently than calling `.half()` did before, though, somehow, maybe it could be helpful.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -362,79 +750,116 @@ def __getitem__(self, idx) -> Tuple[torch.Tensor, torch.Tensor]:
...
+elif args.zero_stage == 3 and is_main_process():
+    logger.info("DeepSpeed ZeRO-3 Memory Estimates")
+    estimate_fn = estimate_zero3_model_states_mem_needs_all_live
+if estimate_fn:
+    estimate_fn(model=model,
+                num_nodes=1,
+                num_gpus_per_node=torch.cuda.device_count())
+
+# The latest deepspeed logging is pretty obnoxious, so we disable it.
+deepspeed.utils.logger.setLevel(logging.WARNING)
```
Eta0, line 793: The `deepspeed` logger is a bit obnoxious but it does also provide a lot of useful information during initialization, enough that I found myself re-enabling it for most of the work I was doing on this script. It might be good to control this behaviour through a command line flag like the regular logging level, or at least to re-enable this whenever `args.log_level.upper() == "DEBUG"`.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -4,37 +4,53 @@
...
-except ModuleNotFoundError:
-    pass
+from typing import Tuple, List
+import socket
+from contextlib import closing, contextmanager
+import logging
 import validators
+import deepspeed
+from deepspeed.runtime.zero.stage_1_and_2 import estimate_zero2_model_states_mem_needs_all_live;
+from deepspeed.runtime.zero.stage3 import estimate_zero3_model_states_mem_needs_all_live;
```
Eta0, line 29: Sneaky semicolons
```suggestion
from deepspeed.runtime.zero.stage_1_and_2 import estimate_zero2_model_states_mem_needs_all_live
from deepspeed.runtime.zero.stage3 import estimate_zero3_model_states_mem_needs_all_live
```

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -46,61 +62,286 @@
...
+args.no_resume = args.no_resume.lower() not in FALSE
+
+# Discover if we have any checkpoints to resume from.
+if not args.no_resume:
+    try:
+        output_dir_list = os.listdir(output_dir)
+        lastCheckpoint = sorted(
+            output_dir_list, key=lambda x: int(x.split("-")[1])
+        )[-1]
+    except:
```
Eta0, line 268: A bare `except:` swallows unusual things like `SystemExit` and `KeyboardInterrupt`, so `except Exception:` ought to be used in parts like this if there is no more specific exception type known.
```suggestion
    except Exception:
```

finetuner-workflow/finetuner/requirements.txt
```diff
@@ -1,13 +1,7 @@
...
-numpy~=1.22.3
-pynvml~=11.4.0
-wandb~=0.12.10
-torch~=1.13.0
-psutil~=5.9.0
-validators
-requests
-accelerate~=0.14.0
+transformers~=4.27.2
+deepspeed==0.8.3
```
harubaru, line 2: This should be updated to 0.9.0 for PyTorch 2.0 support, since certain version asserts in 0.8.3 break with PyTorch 2.0.

wbrown: Done. Let's see if this works.

finetuner-workflow/finetune-workflow.yaml
Eta0: Addressed somewhat in #186. The workflow's `retokenize` parameter is now correctly passed to the tokenizer, and defaults to `'true'` instead of `'false'`.

I agree that not retokenizing is spicy with the tokenizer's current false-negative rate when detecting changes. If more detailed metadata were saved alongside the tokenized dataset (e.g. in a sidecar `.json` file with the same name prefix) and used to check for differences, I think it would be safe to switch `retokenize` back to a default of `'false'`. Until then, I'd leave it as an optional optimization chosen by the workflow user.

finetuner-workflow/finetuner/finetuner.py
Eta0: #186 fixes this to respect `FORCE_FP16` / `--fp16`, but the concern about whether or not to downcast the model at all at this step could still use some investigation.

NavarrePratt (changes requested):
Should we change the default region from `ORD1`. I don't think this comment is true anymore:

> CoreWeave region to default to; ORD1 has most of the GPUs.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -109,34 +151,50 @@ spec:
...
     # Container images -- generally, don't alter this.
     - name: model_downloader_image
       value: 'ghcr.io/wbrown/gpt_bpe/model_downloader'
     - name: model_downloader_tag
-      value: 'e2ef65f'
+      value: 'fd6ec51'
+    - name: tokenizer_image
+      value: 'ghcr.io/wbrown/gpt_bpe/dataset_tokenizer'
+    - name: tokenizer_tag
+      value: 'fd6ec51'
```
NavarrePratt, line 174: This image should be updated to the newest when [this PR](https://github.com/wbrown/gpt_bpe/pull/34) gets merged.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -206,11 +272,25 @@ spec:
               --project_id {{workflow.parameters.project_id}}
               --epochs {{workflow.parameters.epochs}}
               --context_size {{workflow.parameters.context}}
+              --prompt_file {{=workflow.parameters.prompt_file == '' ? '' : '/' + workflow.parameters.pvc + '/' + workflow.parameters.prompt_file}}
```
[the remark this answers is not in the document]
Eta0: This is fixed in #195 by using `=` as an argument delimiter rather than a space, which correctly handles blank arguments (so just `--prompt-file=` works for specifying an empty string). Nice catch, thank you.

harubaru (changes requested):

PR #130, opened 2023-01-23
feat(vso): Update examples

goock:
The PR upgrade examples for Virtual Server.
All examples were tested.

@microcosem I don't think we need to update anything in the doc but just let you know I'm updating the examples.

### DIFF
--- a/virtual-server/examples/curl/run.sh
+++ b/virtual-server/examples/curl/run.sh
@@ -183,9 +183,6 @@ VS_NAME=vs-example
 echo -e "\n------------- CREATE VS -----------------------------------"
 create_vs ${VS_NAME}
 
-echo -e "\n------------- START VS -----------------------------------"
-start_vm ${VS_NAME}
-
 echo -e "\n------------- WAIT FOR VMI UNTIL RUNNING ------------------"
 wait_until_vs_status ${VS_NAME} expect_vs_running
 
--- a/virtual-server/examples/python/kubevirtclient.py
+++ b/virtual-server/examples/python/kubevirtclient.py
@@ -1,4 +1,4 @@
-import os
+import time
 import six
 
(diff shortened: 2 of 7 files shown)


virtual-server/examples/curl/run.sh
```diff
@@ -183,9 +183,6 @@ VS_NAME=vs-example
 echo -e "\n------------- CREATE VS -----------------------------------"
 create_vs ${VS_NAME}
 
-echo -e "\n------------- START VS -----------------------------------"
```
goock, line 186: I simplified the script and removed the start, we have one later.

virtual-server/examples/python/kubevirtclient.py
```diff
@@ -242,7 +242,12 @@ def start(self, namespace, name, **kwargs):
         """
         kwargs['_return_http_data_only'] = True
         version = self.version(namespace, name)
-        return self.kubevirt_api(namespace, name, 'start', version=version, **kwargs)
+        while True:
```
goock, line 245: After the stop command, sometimes VMI is still running and returns 409.

virtual-server/examples/python/main.py
```diff
@@ -104,22 +104,19 @@
 print(vsclient.create(my_virtualserver))
 print(f'VirtualServer status: {vsclient.ready(namespace, name)}')
 
+# Stop the Virtual Machine Instance to apply changes.
```
goock, line 107: Same, I simplified the script, we have another stop later.

virtual-server/examples/python/kubevirtclient.py
YitzyD: Only really if the VM is quickly started/stopped. Good check nonetheless

YitzyD (approved):

PR #132, opened 2023-01-24
Prevent PyTorch from getting overwritten if already installed

harubaru:


wbrown (changes requested):
This won't quite do it  Use `~=` instead.  Could be wrong, though.

harubaru:
Using `~=` has pip replace the installed torch version since it only checks if the package matches the major and minor version but not the patch version, so using `>=` makes sure it doesn't get replaced.

wbrown (approved):

PR #133, opened 2023-01-24
fix(finetuner): Update dataset downloader perms

NavarrePratt:
Part 2 of the fix for [Issue #2 on the dataset-downlaoder](https://github.com/coreweave/dataset-downloader/issues/2).

The dataset downloader image runs as the `nonroot` user so an init container needs to be used to update the mounted PVC's dataset folder to allow others to read/write.


finetuner-workflow/finetune-download-dataset.yaml
```diff
@@ -5,9 +5,23 @@ metadata:
...
+      # needs the correct permissions.
+      initContainers:
+        - name: init
+          image: alpine:3.17
+          command: [ "/bin/sh" ]
+          args:
+            - -c
+            - >-
+              mkdir -p /mnt/pvc/dataset && 
+              chmod o+rw,g+s /mnt/pvc/dataset
```
wbrown, line 18: this should probably be templated as part of the argo job, as `dataset` is a parameter in the workflow.

wbrown: Consider making this part of the argo workflow.

wbrown (changes requested):

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -242,6 +254,50 @@ spec:
...
+  - name: dataset-downloader
+    inputs:
+      parameters:
+        - name: output
+    retryStrategy:
+      limit: 1
+    # The dataset downloader runs as the nonroot user so the dataset folder in the PVC
+    # needs the correct permissions.
+    initContainers:
+      - name: dataset-perms
```
NavarrePratt, line 266: fyi, don't name your init container `init` in argo workflows since that overrides the container named `init` that argo adds under the hood to set things up.

NavarrePratt:
This PR should be merged at the same time as the gitbook changes updating the ML Finetuning page go out.

The gitbook changes are being tracked in [this issue[(https://github.com/coreweave/docs/issues/118)

finetuner-workflow/finetune-workflow.yaml
wbrown: lol :)

wbrown (approved):

PR #162, opened 2023-03-01
Update Model-downloader image

Rexwang8:
Updating model-downloader image for the finetuner-workflow from 7ffb3ec to  e2ef65f 

this update allows the model-downloader to support the new metadata version from hugging face, where the merges.txt and vocab.json is included in the tokenizer.json as a key in the json.


Rexwang8:
This 2nd push updates the image for the dataset-downloader image. Previously, it wasn't functional with the argo workflow call but now it will work properly. This uses the smashwords-downloader repo to get an example dataset to test with.

PR #166, opened 2023-03-09
feat(dreambooth): Add dreambooth to sd finetuner

NavarrePratt:
This PR adds support for the [Dreambooth method](https://dreambooth.github.io/) in the stable diffusion finetuner. It also adds an example Argo workflow that does Dreambooth finetuning and creates an inference service with the finetuned model.

Inspired by [huggingface's dreambooth training script](https://github.com/huggingface/diffusers/tree/main/examples/dreambooth).

**Dreambooth TLDR**
The point of dreambooth finetuning is to teach the model the meaning of a specific object/face with little training data. The same prompt will be used for all instance images and it will contain "unique text" that will be used to describe the new object. Once finetuned, you can use the unique text to generate images with your specific object/face. For example, given 5 pictures of your dog, you can use the prompt "a photo of sks dog". Then during inference you can use "sks dog" in the prompt and the model should generate images with your dog. To improve performance, we also implement prior preservation loss which attempts to make sure the model doesn't lose it's understanding of the generic object (all other dogs).

**Datasets**
The dataset structure for dreambooth differes from the normal text inversion finetuning. Instead of a single folder full of image and text file pairs, there are two folders of image files. One folder for your instance images (photos of _your_ dog) and the other is for class images (photos of dogs). The same prompt is for all instance images, and for all class images. The dataset classes are all implemented in `datasets.py`.

**Training loop**
Before the training loop starts, enough class images will be generated by the pretrained model. Then during each step of training, both an instance image and class image go through the model simultaneous and used to calculate the prior preservation loss. 

**Argo Workflow**
The new dreambooth argo workflow is very similar to the existing stable diffusion finetuner. The main differences are the different parameters to support the new finetuning method.

**Testing**
To test these changes, I've run both the finetuning Argo workflows all the way through.

**Envisioned Next Steps**
- ~~Use tensorizer for serializing and deserializing the model checkpoints~~ Done
- Trigger Argo workflow run when new instance image folder is created in a S3 bucket

### DIFF
--- a/online-inference/stable-diffusion/Dockerfile
+++ b/online-inference/stable-diffusion/Dockerfile
@@ -2,19 +2,14 @@ ARG CUDA_RELEASE=11.6.2-cudnn8-devel-ubuntu20.04
 FROM nvidia/cuda:${CUDA_RELEASE} AS base
 ENV DEBIAN_FRONTEND=noninteractive
 ENV tenzorizer_commit=35381e3812ba342991d30b71ce257503622ae828
+
+RUN apt update && apt upgrade -y && \
+    apt update && apt install -y python3 python3-pip git curl \
+
 ADD service/ /app/
 WORKDIR /app
 
-RUN apt update && apt upgrade -y && \
-    apt update && apt install -y python3 python3-pip git curl && \
-    pip3 install --no-cache-dir --upgrade pip && \
+RUN pip3 install --no-cache-dir --upgrade pip && \
     pip3 install --no-cache-dir -r requirements.txt
-
-RUN git clone https://github.com/coreweave/tensorizer && \
-    cd tensorizer && \
-    git checkout ${tenzorizer_commit} && \
-    cd .. && \
-    mv tensorizer/tensorizer.py . && \
-    rm -rf tensorizer
 
 CMD ["python3", "/app/service.py"]
--- a/online-inference/stable-diffusion/Dockerfile
+++ b/online-inference/stable-diffusion/Dockerfile
@@ -1,10 +1,10 @@
 ARG CUDA_RELEASE=11.6.2-cudnn8-devel-ubuntu20.04
 FROM nvidia/cuda:${CUDA_RELEASE} AS base
 ENV DEBIAN_FRONTEND=noninteractive
-ENV tenzorizer_commit=35381e3812ba342991d30b71ce257503622ae828
 
 RUN apt update && apt upgrade -y && \
-    apt update && apt install -y python3 python3-pip git curl
+    apt update && apt install -y python3 python3-pip git curl && \
+    apt clean
 
 ADD service/ /app/
 WORKDIR /app
--- a/online-inference/stable-diffusion/Dockerfile
+++ b/online-inference/stable-diffusion/Dockerfile
@@ -4,7 +4,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 ENV tenzorizer_commit=35381e3812ba342991d30b71ce257503622ae828
 
 RUN apt update && apt upgrade -y && \
-    apt update && apt install -y python3 python3-pip git curl \
+    apt update && apt install -y python3 python3-pip git curl
 
 ADD service/ /app/
 WORKDIR /app
--- a/sd-dreambooth-workflow/db-workflow-event-binding.yaml
+++ b/sd-dreambooth-workflow/db-workflow-event-binding.yaml
@@ -1,0 +1,39 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowEventBinding
+metadata:
+  name: db-finetune-event-binding
+spec:
+  event:
+    selector: discriminator == "db-finetune"
+  submit:
+    workflowTemplateRef:
+      name: db-finetune-template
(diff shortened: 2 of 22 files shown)


sd-finetuner-workflow/sd-finetuner/finetuner.py
```diff
@@ -232,128 +294,6 @@ def get_gpu_ram() -> str:
...
-                    truncation=True,
-                ).input_ids
-        except Exception as e:
-            print(
-                f'Error with {self.examples[self.hashes[i]]["image"]} -- {e} -- skipping {i}'
-            )
-            return None
-
-        if random.random() < self.ucg:
-            image["caption"] = ""
```
NavarrePratt, line 352: I've removed UCG as an argument in my changes because I noticed it wasn't properly implemented. This if statement sets a `caption` value to an empty string but the key that is used everywhere else in the code is `input_ids`.

@harubaru - Is just emptying the `input_ids` enough to drop text conditioning or would we need extra logic around it in the step function when using the UNet model?

salanki:
The customers here want to programatically kick off a finetune and get results, potentially even running an inference  against it. We should have an end to end example on how to programatically kick off an Argo Workflow (using a workflowtemplate)? And storing the resulting images/model in CW object storage maybe?

NavarrePratt:
> The customers here want to programatically kick off a finetune and get results, potentially even running an inference against it. We should have an end to end example on how to programatically kick off an Argo Workflow (using a workflowtemplate)? And storing the resulting images/model in CW object storage maybe?

@salanki - Yeah I agree, there's a lot more we should do to beef up the argo workflow example. I've briefly looked into [S3 sensors](https://argoproj.github.io/argo-events/tutorials/03-trigger-sources/#s3) so the customer can have whatever frontend/middleware drop the instance images in some bucket which automatically trigger a new finetune and inference service. Whatever we do, I'd like to have an example showing how you can leverage deploying many scale-to-0 inference services before we write this up into a docs page.

Also waiting on [this tensorizer PR](https://github.com/coreweave/tensorizer/pull/12) before I add tensorizer support to this. I wanted to get the finetuner code reviewed and merged so the PRs don't get too big.

salanki:
Triggering on S3 is a cool option for sure, but it does make sense to document how to trigger one with an Argo webhook, especially since we don't have a managed Argo Events today.

sd-finetuner-workflow/sd-finetuner/finetuner.py
harubaru: I cross-referenced the code based off of the original implementation which also just dropped text conditioning only:
https://github.com/pesser/stable-diffusion/blob/57eea7dfc2cdd8cadae77ab1c391f956d46f69bd/ldm/models/diffusion/ddpm.py#L396

geekyayush (approved):
LGTM

sd-dreambooth-workflow/db-workflow-event-binding.yaml
```diff
@@ -0,0 +1,39 @@
...
+            event: payload.output
+        - name: num_class_images
+          valueFrom:
+            event: "payload.num_class_images == null ? 100: payload.num_class_images"
+        - name: run_inference
+          valueFrom:
+            event: "payload.run_inference == null ? true : payload.run_inference"
+        - name: inference_only
+          valueFrom:
+            event: "payload.inference_only == null ? false : payload.inference_only"
```
NavarrePratt, line 39: Once the `WorkflowTemplate` and `WorkflowEventBinding` are deployed, you can trigger a workflow run with a webhook like the following:
```bash
curl --location 'https://<Argo deployment>/api/v1/events/<namespace>/db-finetune' \
--header 'Authorization: Bearer <REDACTED>' \
--header 'Content-Type: application/json' \
--data '{
    "run_name": "event-test",
    "instance_dataset": "data/example-dog",
    "instance_prompt": "a photo of sks dog",
    "class_dataset": "generic/dogs-2",
    "class_prompt": "a photo of dog",
    "num_class_image": 10,
    "output": "finetunes/event-test"
}'
```

online-inference/stable-diffusion/Dockerfile
```diff
@@ -2,19 +2,14 @@ ARG CUDA_RELEASE=11.6.2-cudnn8-devel-ubuntu20.04
 FROM nvidia/cuda:${CUDA_RELEASE} AS base
```
wbrown, line 2: For later, but we may want to base this off our `torch-base` image.

wbrown (approved):
Some minor comments, nice job. :+1:

online-inference/stable-diffusion/Dockerfile
```diff
@@ -2,19 +2,14 @@ ARG CUDA_RELEASE=11.6.2-cudnn8-devel-ubuntu20.04
...
 ENV DEBIAN_FRONTEND=noninteractive
 ENV tenzorizer_commit=35381e3812ba342991d30b71ce257503622ae828
```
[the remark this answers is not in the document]
NavarrePratt: Correct. The tensorizer versioning is now in the requirements bc we don't need to install from source.

PR #174, opened 2023-03-31
Implement Tensorizer Model Loading for GPT Finetuner

harubaru:
Implemented an optional `tensorizer_uri` parameter that allows model weights to be downloaded from either a PVC or from S3. This allows for a much-needed decrease in initial system memory usage which could cause a pod to run out of memory in some instances.


finetuner-workflow/finetune-workflow.yaml
```diff
@@ -251,6 +263,22 @@ spec:
...
+  - name: check-model
+    inputs:
+      parameters:
+        - name: model
+    retryStrategy:
+      limit: 1
+    container:
+      image: "alpine:3.12"
+      command: [ "/bin/sh", "-c" ]
+      args: [ "curl -I -XGET -s -o /dev/null -w \"%{http_code}\" https://accel-object.ord1.coreweave.com/tensorized/{{inputs.parameters.model}}/model.tensors | grep -q 200 && echo -n 'true' > /tmp/output.txt || echo -n 'false' > /tmp/output.txt" ]
```
wbrown, line 275: Does a `HEAD` not work instead of a `GET`? You're fetching the entire file here.

wbrown (changes requested):

finetuner-workflow/finetune-workflow.yaml
harubaru: The `-I` argument should be getting just fetches the header information. Running this takes a second to execute since it only fetches the header:
`curl -I -XGET -s -o /dev/null -w \"%{http_code}\" https://accel-object.ord1.coreweave.com/tensorized/EleutherAI/pythia-2.8b/model.tensors`
Whereas running the same command without the `-I` argument takes quite a lot longer, since it is actually fetching the whole file:
`curl -XGET -s -o /dev/null -w \"%{http_code}\" https://accel-object.ord1.coreweave.com/tensorized/EleutherAI/pythia-2.8b/model.tensors`
A more readable flag would be `--head` though lol.

wbrown (approved):

PR #175, opened 2023-04-08
LLM Finetuner Update Polish

Eta0:
## Finetuner Update Polish
This PR contains some fixes, improvements, and refactoring to polish off the LLM finetuner update (#128).

### Dataset Splitting (Bug)
When the `no_shuffle` flag was enabled, the finetuner did not correctly respect the train/eval ratio.
It would always use 100% of the dataset for training, and overlap that with the (shuffled, ironically) eval dataset, even though those shouldn't be mixed. This PR changes the implementation to extract a contiguous block (placed randomly) for the eval dataset, and to use the rest of the dataset minus that particular section for the train dataset, all while preserving the original order.

### DeepSpeed Logging (Feature)
DeepSpeed logging is now re-enabled whenever the `--log-level` command line argument to the finetuner is set to `DEBUG`, for maximal information.

### Dependency Pruning
Several of the listed (and even imported) dependencies for the finetuner were outdated and unnecessary, and have been pruned. This should make it easier for us to keep the remaining ones up to date in the future.

### Reducing Drift Between the Finetuner & Evaluator (Refactor + Improvements)
The `evaluator.py` code is largely a subset of the finetuner code. Because of that, it has a lot of repeated function definitions that were originally copied from the finetuner, but which have not kept pace with the finetuner's updates.

Most near-identical chunks of code between the two files have now been updated to the more recent of the two. For example, the evaluator now supports padding and end-of-sentence tokens in the same way that the finetuner does.

This PR also factors out two utility functions shared between the finetuner and the evaluator into a common `utils.py` file: the memory usage check `get_gpu_ram` and `no_init`.

`get_gpu_ram` has been broken down into four granular, reusable classes, `GlobalGPUMemoryUsage`, `TorchGPUMemoryUsage`, `CPUMemoryUsage`, and the collective `MemoryUsage`, which effectively replace all code that had previously queried memory information—e.g. the implementation of `estimate_batch_size`—in an efficient and (hopefully) readable way. It also completely removes the finetuner's direct dependency on `pynvml`, which is now removed from `requirements.txt` as well.

The evaluator has also been given more descriptive error messages when provided invalid arguments, and its code style (method call indentation, single vs. double quotes) has been synced up with the finetuner.

Additionally, the evaluator now always reads prompt files as UTF-8, like the finetuner did, and the finetuner now handles CRLF (and CR) linefeeds in prompt files, like the evaluator did. Sharing is caring!

#### Odd Remaining Bits
At the moment, the evaluator still contains this code:
```py
if args.fp16:
    model = model.half().to(device)
else:
    model = model.to(device)
```
Even though that had been explicitly removed from the finetuner in a previous commit, as it may be more helpful (and correct) in the evaluator in ways that it wasn't for the finetuner. I didn't benchmark it to find out. The datatype casting in both files could likely still use a bit of scrutiny.

### DIFF
--- a/finetuner-workflow/finetuner/evaluator.py
+++ b/finetuner-workflow/finetuner/evaluator.py
@@ -118,7 +118,6 @@ model.eval()
 
 duration = time.time() - start
 print(f"Loaded model in {duration:.2f}s")
-print(MemoryUsage.now())
 torch.cuda.memory.empty_cache()
 print(MemoryUsage.now())
 
--- a/finetuner-workflow/finetuner/evaluator.py
+++ b/finetuner-workflow/finetuner/evaluator.py
@@ -138,9 +138,9 @@ def evaluate(
     """
     output_texts: List[str] = []
     input_tokens: Tensor = (
-        torch.LongTensor(tokenizer.encode(prompt)).unsqueeze(0).to(device)
+        torch.LongTensor(eval_tokenizer.encode(prompt)).unsqueeze(0).to(device)
     )
-    attention_mask: Tensor = input_tokens != tokenizer.pad_token_id
+    attention_mask: Tensor = input_tokens != eval_tokenizer.pad_token_id
     max_length = input_tokens.shape[1] + generate_tokens
 
     generated_tokens = eval_model.generate(
--- a/finetuner-workflow/finetuner/evaluator.py
+++ b/finetuner-workflow/finetuner/evaluator.py
@@ -1,19 +1,13 @@
 #!/bin/which python3
 import argparse
-from typing import Callable, List
+from typing import List
 from torch import Tensor
-import resource
-import pynvml
-import psutil
 import time
 import torch
 from transformers import AutoTokenizer, AutoModelForCausalLM, PreTrainedTokenizer
-from transformers.modeling_utils import no_init_weights, PreTrainedModel
+from transformers.modeling_utils import PreTrainedModel
 
-try:
-    pynvml.nvmlInit()
-except pynvml.nvml.NVMLError_LibraryNotFound:
-    pynvml = None
+from utils import *
 
 device = torch.device("cpu")
 if torch.cuda.is_available():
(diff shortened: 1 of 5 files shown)


finetuner-workflow/finetuner/evaluator.py
```diff
@@ -101,101 +53,113 @@ def get_gpu_ram() -> str:
...
+    prompt,
+    generate_tokens: int = 200,
+    num_samples: int = 1,
+    eval_model: PreTrainedModel = model,
+    eval_tokenizer: PreTrainedTokenizer = tokenizer,
+    top_k: int = args.top_k,
+    top_p: float = args.top_p,
+    temperature: float = args.temperature,
+    repetition_penalty: float = args.repetition_penalty
+) -> List[str]:
```
NavarrePratt, line 136: To prevent drift going forward, should we just have the fintuner import `evaluate` from `evaluator.py`? Or at least have the `evaluate` function somewhere common between the two?

We'd have to move all the global level stuff in this script into a `main` function so it doesn't get run when importing `evaluate`, but I think that's better then making sure the functions across both files always match up during changes.

Thoughts?

finetuner-workflow/finetuner/evaluator.py
```diff
@@ -101,101 +53,113 @@ def get_gpu_ram() -> str:
...
-def evaluate(prompt,
-             generate_tokens: int = 200,
-             num_samples: int = 1,
-             eval_model: PreTrainedModel = model,
-             eval_tokenizer: PreTrainedTokenizer = tokenizer,
-             top_k: int = args.top_k,
-             top_p: float = args.top_p,
-             temperature: float = args.temperature,
-             repetition_penalty: float = args.repetition_penalty) -> List[str]:
+print(MemoryUsage.now())
```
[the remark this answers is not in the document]
Eta0: I'm not sure. It was like that before the refactor, but the first one may have been extraneous. The second one is probably more relevant.

finetuner-workflow/finetuner/evaluator.py
Eta0: This was something I was already considering during this round of changes and here are my thoughts on it:

- On one hand, we could factor out nearly everything from `evaluator.py` except for the CLI—the model loading, the tokenizer setup, the `evaluate` function, and even the prompt file parsing are all shared with the finetuner. It would make the collection of the finetuner and evaluator scripts cleaner as a pair to share even more between each other.
  - It's also good to keep in mind that the shared model loading code will probably be getting more complex with the upcoming `tensorizer` integration, so separating that part out from the finetuner could just be cleaner in its own right.
- On the other hand, this code also serves as an example in this repository of writing an LLM finetuner and evaluator. Having standalone examples of the relevant ML logic presented step-by-step in each single file *could* potentially be more helpful to people who are only interested in one or the other.

Having the parts that pertain to evaluation be imported to the finetuner straight out of the evaluator also makes sense conceptually IMO. Though it also re-raises the question about whether or not to also deduplicate the model loading code, which conceptually could belong to either.

I don't have a strong opinion either way, so I'd defer to @wbrown for what he thinks would be better.

wbrown (approved):
Nice job. I'm OK with merging this in as is, but you can do the `evaluator.py` factorization if you want.

wbrown:
@Eta0 I'm going to merge this, and if we want to do the `evaluator.py` factorization, we can do it as a separate PR.

Issue #183, opened 2023-04-25
KServe transitive dependency license issue


Eta0:
We currently use [`kserve`](https://github.com/kserve/kserve) (a.k.a. `kfserving` in older versions) as a dependency for the following `online-inference` examples:
- [`bloom-176b`](https://github.com/coreweave/kubernetes-cloud/tree/ad06351de00bc9d8b732f3d95ea5486ea71f1a4f/online-inference/bloom-176b/model/requirements.txt)
- [`custom-pytorch-aitextgen`](https://github.com/coreweave/kubernetes-cloud/tree/ad06351de00bc9d8b732f3d95ea5486ea71f1a4f/online-inference/custom-pytorch-aitextgen/custom-predictor/requirements.txt)
- [`custom-sentiment`](https://github.com/coreweave/kubernetes-cloud/tree/ad06351de00bc9d8b732f3d95ea5486ea71f1a4f/online-inference/custom-sentiment/custom-predictor/requirements.txt)
- [`stable-diffusion/service`](https://github.com/coreweave/kubernetes-cloud/tree/ad06351de00bc9d8b732f3d95ea5486ea71f1a4f/online-inference/stable-diffusion/service/requirements.txt)

[KServe issue #2683](https://github.com/kserve/kserve/issues/2683) notes that their (previous) logging dependency [`table-logger`](https://github.com/AleksTk/table-logger) is exclusively GPL-licensed in all of its PyPI releases.

- `table-logger` switched to an MIT license [in a commit in February](https://github.com/AleksTk/table-logger/commit/3cab68c59062ce4d9cb3b667f64e83a23789bf63), with the rest of its codebase in an identical state to its latest PyPI release (0.3.6). Nonetheless, the PyPI distribution that we use technically only embeds the GPLv2 license. They have not released anything on PyPI with the updated license.
- `kserve` moved away from `table-logger` entirely in favour of [`tabulate`](https://pypi.org/project/tabulate/) (MIT-licensed) in `kserve` release `0.10.1`.

Since this repository is not GPL-licensed, for compliance, we should either:
- Pin the `table-logger` transitive dependency in our `requirements.txt` files alongside older `kserve`/`kfserving` dependencies to the latest Git version (`git+https://github.com/AleksTk/table-logger@3cab68c59062ce4d9cb3b667f64e83a23789bf63`), which embeds the MIT license, or
- Update `kserve` dependencies to at least version `0.10.1`

[...]

PR #184, opened 2023-04-25
Fix `kserve` transitive dependency licensing issues

Eta0:
Start of a fix for #183. Docker images still need to be rebuilt and re-published with updated dependencies.

This adds an extra requirement of `-e git+https://github.com/AleksTk/table-logger@3cab68c59062ce4d9cb3b667f64e83a23789bf63#egg=table-logger` alongside older `kserve` and `kfserving` dependencies in the `online-inference` examples.
- There is no PyPI release available with the updated license, so this installs the first Git version of the library with the updated license. It is otherwise identical to the latest PyPI release of `table-logger` (version 0.3.6).
- `-e` (editable mode) is necessary with the Git installation because otherwise the MIT license file isn't properly bundled.

### DIFF
--- a/online-inference/bloom-176b/model/requirements.txt
+++ b/online-inference/bloom-176b/model/requirements.txt
@@ -1,3 +1,6 @@
 accelerate==0.11.0
 git+https://github.com/huggingface/transformers.git@ccc0897 # For device_map in pipelines() support
 kserve==0.8.0.2
+
+# MIT-licensed version of a kserve transitive dependency (necessary for kserve < 0.10.1)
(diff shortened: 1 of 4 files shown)


wbrown (approved):
:+1:

wbrown:
Screenshot 2023-05-02 at 2 33 09 PM
@Eta0 Unfortunately, `snyk` does not appear to be appeased.

Eta0:
> @Eta0 Unfortunately, `snyk` does not appear to be appeased.

A plague o' their houses. I can wrangle up an ignore file for it to add here.

PR #185, opened 2023-04-25
feat(spark): Add a spark example

NavarrePratt:
https://github.com/coreweave/docs/issues/228

This PR adds examples of running Spark on Kubernetes. Both of the examples use [img2dataset](https://github.com/rom1504/img2dataset).

The first example runs the spark job in "cluster" mode, and creates the job using `spark-submit`. This is the "simple" example and downloads a small dataset with 1 executor.

The second example involves deploying a service running Jupyter Lab. Then the spark job can be run from within a notebook in client mode. This example uses dynamic executors to scale up and down based on the current jobs, and will use 5 executors to download CC12M in ~1 hour. It includes logging to Weights and Biases.

This PR will stay as a draft until the docs are ready.

TODO Before merge:
- [x] Create docs page
- [x] Confirm security of JupyterLab deployment

### DIFF
--- a/spark/jupyter/interactive-example.ipynb
+++ b/spark/jupyter/interactive-example.ipynb
@@ -1,0 +1,324 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9ee8ceb8-e788-4fd4-97f4-aac3c79e41cf",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "!pip install pyspark img2dataset"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7a32f009-a31b-4b8e-8109-ccae087ab4d3",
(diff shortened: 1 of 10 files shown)


spark/jupyter/interactive-example.ipynb
```diff
@@ -0,0 +1,324 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "76d6e604-a3f7-4310-9a85-a1022d2bd0c1",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# TODO: Update with your namespace\n",
+    "NAMESPACE = \"tenant-sta-nav-npratt\""
```
wbrown, line 37: You may want to change this to `tenant-CHANGEME` :)

wbrown (approved):
This looks great. Not fond of how `.ipynb` renders Python code and how they show up in diffs, but not much we can do about it.

One comment you may want to address.

PR #186, opened 2023-04-28
More finetuner update fixes.

Eta0:
# Finetuner fixes and improvements (for #128).
- f23b16dbac077b2134f849a3ab36a3a67b0803d7: Fixed `PerformanceCallback` accidentally being disabled when no `--prompt_file` is given.
- 2f677311320502677f4c6b74ec8c0e98523aa993: Replaced manual GPU-counting with DeepSpeed's world size when calculating the number of contexts trained.
- e37b7d01dd93df1e862b0039dfd671627a23226d: Temporarily disabled `train_ratio` as a workflow parameter.
  - It isn't implemented correctly as-is, and simply discards training data with no benefit.
  - For now, `finetuner.py` emits a warning if it is passed a `--train_ratio` other than `1.0`, and overrides it to `1.0`.
- 9a06e683cb7731b3767d97f485d673692851521f: Stopped masking end-of-sentence tokens as padding if the finetuner can't tell the two apart.
  - This was preventing the finetuner from learning the EOS token and respecting the boundaries between texts under the default settings.
  - As an exception, the final dataset context will still have padding tokens masked even if they are indistinguishable from EOS, since `gpt_bpe` currently only puts padding tokens in the final context (`-reorder=shuffle` notwithstanding).
  - This is still only a workaround; a better solution is to make `gpt_bpe` emit unambiguous token masks at tokenization time.
- 9f0318e8345bb8a860aed938a7c457fbb1b871b3: Fix the workflow's `retokenize` parameter (which did nothing), and expose the tokenizer `-sanitize` flag as a workflow parameter.
  - This additionally changes the default for `retokenize` to `'true'`.
    - The potential confusion that would arise from using an incorrectly tokenized dataset likely outweighs the benefit of optimizing tokenization passes.
    - This way, the optimization can still be enabled when someone knows what they're doing.

### DIFF
--- a/finetuner-workflow/finetune-workflow.yaml
+++ b/finetuner-workflow/finetune-workflow.yaml
@@ -24,7 +24,11 @@ spec:
     # Whether to retokenize the data on each run. It is strongly recommended to
     # NOT enable this if you are making multiple runs on the same dataset.
     - name: retokenize
-      value: 'false'
+      value: 'true'
+    # Whether to sanitize the data during tokenization,
+    # fixing common whitespace and line-ending issues automatically.
+    - name: sanitize
+      value: 'true'
     # Which tokenizer to use - the dataset_tokenizer internally supports `gpt2`
     # and `pile`. It defaults to the model's tokenizer data if not provided.
     - name: tokenizer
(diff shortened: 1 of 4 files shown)


wbrown (approved):
Excellent updates, thank you!

PR #187, opened 2023-05-01
CURL fixes for GPT finetuner

harubaru:
`alpine:3.12` apparently did not have CURL installed, replacing with `alpine/curl:3.14` and using an absolute path to the CURL binary fixes this issue where CURL is not found.

### DIFF
--- a/finetuner-workflow/finetune-role.yaml
+++ b/finetuner-workflow/finetune-role.yaml
@@ -1,13 +1,19 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: inference
+  name: finetune
 ---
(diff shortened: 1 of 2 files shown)


NavarrePratt:
Instead of having to pass the name of the service account we want to use whenever it gets triggered, we can add the service account right to the workflow.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: "finetune-"
spec:
  entrypoint: main
  serviceAccountName: finetune
  arguments:
...
```

wbrown:
@NavarrePratt Good point!
@harubaru let's do that.

PR #189, opened 2023-05-01
Finetuner fixes: Random seeding, better empty model loading, more code cleanup

Eta0:
More fixes and improvements for #128.
- 38c5e7e75438c0506f7165176f7497ec7f393b54: Removed the unused `find_free_port` function, since it can't be used with the DeepSpeed launcher.
- f40f0b908b458e4cd7209b6bf72a139fa36c710f: Reorganized `TrainingArguments` constructor arguments and some minor improvements in setting `do_train`, `model.train()`, commenting out unused arguments, etc.
- 1f8e06df01723ed8e852411fb4989c9e850dd3e3: Set the finetuner's seed in the `TrainingArguments` constructor so it will actually be used during finetuning, and not be overwritten.
- 7d7438afa1bf52ce9c40a476425691cf596901dd: Changed initialization of the empty model for the tensorizer loading code to use `AutoModelForCausalLM.from_config()` instead of `AutoModelForCausalLM.from_pretrained()`
  - `AutoModelForCausalLM.from_config()` is specifically designed to be used when creating a model without loading any specific weights, whereas `AutoModelForCausalLM.from_pretrained()` prints a massive warning when used that way.

If we need the `find_free_port()` function later, since it must be used separately among the DeepSpeed launcher's command line arguments, it should be re-added as its own standalone script:
```py
#!/bin/which python3
from contextlib import closing
import socket


def find_free_port():
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
        s.bind(("", 0))
        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        return s.getsockname()[1]


if __name__ == "__main__":
    print(find_free_port())
```

### DIFF
--- a/finetuner-workflow/finetuner/finetuner.py
+++ b/finetuner-workflow/finetuner/finetuner.py
@@ -20,7 +20,6 @@ import argparse
 import pathlib
 from typing import Tuple, List
 import socket
-from contextlib import closing
 import logging
 import deepspeed
 from deepspeed.runtime.zero.stage_1_and_2 import (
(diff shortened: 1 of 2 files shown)


wbrown (approved):
Good changes, thank you. :+1:

PR #195, opened 2023-05-06
build(finetuner): Pre-compile CUDA dependencies in another build stage

Eta0:
Fix for building the finetuner image in #128.

This separates out compilation of DeepSpeed ops and `flash-attn` into a separate Dockerfile build stage so that their heavy build dependencies (NVCC + ~2 GiB of libraries) don't need to be bundled in the finetuner image. This also fixes building `flash-attn`, which was slightly broken.

These use the `TORCH_CUDA_ARCH_LIST` environment variable inherited from the `torch:base` image to automatically compile for all the architectures supported by the `torch` installation.

Only two DeepSpeed ops are built, specified by `DS_BUILD_UTILS=1 DS_BUILD_CPU_ADAM=1`. There are [more ops in DeepSpeed](https://www.deepspeed.ai/tutorials/advanced-install/), but only these two are used by the finetuner. Building all ops is more complicated and requires extra dependencies (for `DS_BUILD_AIO` and `DS_BUILD_SPARSE_ATTN`, for example). Nonetheless, more flags can be enabled in the Dockerfile if we need them.


wbrown (approved):
Great job. :+1:

PR #199, opened 2023-05-11
Updated Inference Server for GPT Finetuner

harubaru:
The previous GPT inference server used in the finetuning example was out of date to the point that it cannot run inference on Pythia or GPT-NeoX models. The changes in this PR also incorporates an inference server with the finetuner image itself for simplicity.

Sidenote, #198 was made out of accident due to me being a little sleep deprived :sweat_smile:

### DIFF
--- a/finetuner-workflow/finetuner/Dockerfile
+++ b/finetuner-workflow/finetuner/Dockerfile
@@ -12,5 +12,6 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 COPY ds_config.json .
 COPY finetuner.py .
 COPY evaluator.py .
+COPY inference.py .
 COPY utils.py .
 CMD [ "/usr/bin/python3", "finetuner.py" ]
--- a/finetuner-workflow/finetuner/requirements.txt
+++ b/finetuner-workflow/finetuner/requirements.txt
@@ -6,6 +6,6 @@ torch==2.0.0
 psutil==5.9.4
 accelerate~=0.17.1
 tensorizer==1.1.0
-einops
-flash-attn
+einops==0.6.1
+flash-attn==1.0.4
 kserve==0.10.1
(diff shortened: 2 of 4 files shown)


finetuner-workflow/finetuner/requirements.txt
```diff
@@ -8,3 +8,4 @@ accelerate~=0.17.1
 tensorizer==1.0.1
 einops
 flash-attn
+kserve
```
NavarrePratt, line 11: Few things:

1. We should include some sort of versioning for all the packages. That way everytime a customer builds an image going through the docs, we know it'll be consistent (as much as we can control at least)

2. The pip install fails both locally and in the docker build. Were you able to build the image and test it in the workflow? In the workflow it's still using the goose finetuner image.

3. Should we be bumping tensorizer to the newest version, 1.1.0?

finetuner-workflow/finetuner/Dockerfile
```diff
@@ -12,5 +12,6 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 COPY ds_config.json .
 COPY finetuner.py .
 COPY evaluator.py .
+COPY inference.py .
```
NavarrePratt, line 15: I'm fine with doing this to get everything working ASAP, but we should be making the inference and finetuner images separate.

We at least don't need `deepspeed` in the inference image, and there's probably some other unused dependencies as well.

This is also a general huggingface-transformers-kserve inference script. Should this live somewhere separate like `ml-images`? I think we've been in need of putting all our the nested images in kubernetes-cloud someone else more visible.

harubaru:
For reference, this PR is to be added on top of https://github.com/coreweave/kubernetes-cloud/pull/128

wbrown (approved):
@harubaru I'll approve and merge this, but the workflow should really be passing in the arguments that we have for sampling.

Also, we should be sharing sampling code where possible with `evaluate.py`.

PR #203, opened 2023-05-25
Replace KServe with FastAPI for inference in finetuner example.

harubaru:
This PR replaces KServe's functionality with FastAPI for the sake of simplicity and allowing customers to get a better understanding of how to write an inference service on CoreWeave. Using KServe has a couple downsides, one of which is requiring package versions that conflicts with a key library being used such as [Tensorizer](https://github.com/coreweave/tensorizer).

This PR is to be added on top of #128.

### DIFF
--- a/finetuner-workflow/finetuner/inference.py
+++ b/finetuner-workflow/finetuner/inference.py
@@ -1,16 +1,47 @@
-import os
-import uvicorn
-import torch
+from typing import List, Optional
 
-from pydantic import BaseModel
-from typing import Optional, List
+import torch
+import uvicorn
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
+from pydantic import BaseModel
 from transformers import pipeline
+
+from utils import DashParser
+from utils import validation as val
+
+parser = DashParser(description="Text model inference HTTP server")
+
+parser.add_argument(
+    "--model",
+    type=str,
+    default="distilgpt2",
+    help="Model to use for inference (directory, or HuggingFace ID) [default = distilgpt2]",
+)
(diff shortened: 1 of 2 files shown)


harubaru:
One thing that needs to be done so far is to share sampling code with `evaluate.py`, since this gets inference to _just_ work with FastAPI.

finetuner-workflow/finetuner/inference.py
```diff
@@ -1,71 +1,77 @@
...
-    'TEMPERATURE': float(os.getenv('TEMPERATURE', 1.0)),
-    'TOP_K': int(os.getenv('TOP_K', 50)),
-    'TOP_P': float(os.getenv('TOP_P', 0.95)),
-    'REPETITION_PENALTY': float(os.getenv('REPETITION_PENALTY', 1.0)),
-    'NUM_RETURN_SEQUENCES': int(os.getenv('NUM_RETURN_SEQUENCES', 1)),
+args = {
+    "model": os.getenv("INFERENCE_MODEL", "distilgpt2"),
+    "device": int(os.getenv("INFERENCE_DEVICE", 0)),
+    "port": int(os.getenv("INFERENCE_PORT", 80))
 }
```
Eta0, line 41: IMO using environment variables for this one while all of the other containers use command line arguments seems a little odd. Is there a particular reason to keep them this way, or just convenience?

Eta0 (changes requested):
One required change (`async def` → `def`) and two nits. If you'd like me to whip up an `argparse` CLI for it I'd be happy to contribute one.

Eta0 (approved):
Looks good 👍

PR #204, opened 2023-05-25
feat(update): Adding GPTJ with HF and Tensorizer model loading

rtalaricw:
This PR adds an initial example to load models using Tensorizer


online-inference/tensorizer-isvc/gptj_6b_tensorizer/kserve/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# Not pinned due to tensorizer package resolution
+kserve
```
harubaru, line 2: This should be changed to the version which pip finds to be suitable with `tensorizer`.

online-inference/tensorizer-isvc/gptj_6b_hf/kserve/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# Not pinned due to tensorizer package resolution 
+kserve
```
harubaru, line 2: This should also be changed to the version which pip finds to be suitable with `tensorizer`.

harubaru (changes requested):

online-inference/tensorizer-isvc/benchmark/load_test_async.py
```diff
@@ -0,0 +1,87 @@
...
+    def get(url):
+        return session.get(url).status_code
+    async_get = sync.sync_to_async(get)
+
+    async def get_all(urls):
+        return await asyncio.gather(*[
+            async_get(url) for url in urls
+        ])
+    # call get_all as a sync function to be used in a sync context
+    return sync.async_to_sync(get_all)(urls)
```
Eta0, line 44: This seems to be dead code superseded by `async_aiohttp_get_all`. Should this still be here?

online-inference/tensorizer-isvc/benchmark/load_test_async.py
```diff
@@ -0,0 +1,87 @@
...
+
+inputs = ["Hello, how are you?", "What up dig dog?", "You are a killer!", "JCPenny is a bad store", "Life is great", "Chilling on a roof", "Love you", "Mox is cute", "You are my enemy", "Change is required", "Love the life"]
+
+def main():
+    args = parse_args()
+    urls = []
+    for _ in range(100):
+        urls.append(f"{args.url}/predict/{random.choice(inputs)}")
+    
+    async_aiohttp_get_all(urls)
```
Eta0, line 80: Introducing a dependency on `asgiref` for to call this one function synchronously seems excessive and less clear than using the built-in `asyncio` module itself. You could instead use `asyncio.run` in `main` to run a single `async` function:
```py
def main():
    ...
    timed(asyncio.run)(async_aiohttp_get_all(urls))
```
Or declare `main` as `async`, or (the most general but least necessary option) the `async_main` pattern in base Python:
```py
async def async_main():
    ...  # Can use `await` freely


def main():
    asyncio.run(async_main())


if __name__ == "__main__":
    main()
```

> 🛈 The last option is usually used to have a synchronous entrypoint function to an asynchronous program, e.g. for installable packages with entry points.

online-inference/tensorizer-isvc/benchmark/load_test_sync.py
```diff
@@ -0,0 +1,60 @@
...
+#     @task
+#     def predict(self):
+#         inputs = ["Hello, how are you?", "What up dig dog?", "You are a killer!", "JCPenny is a bad store", "Life is great"]
+
+
+#         with self.client.get(f"/predict/{random.choice(inputs)}") as response:
+#             if response.status_code != 200:
+#                 response.failure("Could not return response")
+
+inputs = ["Hello, how are you?", "What up dig dog?", "You are a killer!", "JCPenny is a bad store", "Life is great", "Chilling on a roof", "Love you", "Mox is cute", "You are my enemy", "Change is required", "Love the life"]
```
Eta0, line 20: JCPenney has a typo in it, but maybe we should avoid picking fights with still-existing companies just for an example string, haha.

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
```diff
@@ -0,0 +1,68 @@
...
+import requests
+import numpy as np
+import base64
+import time
+import torch
+from flask import Flask
+from transformers import AutoTokenizer, GPTJForCausalLM
+
+# The signature name is defined at time of export, in signature_def_map supplied to builder
+# Tensorflows default is serving_default
```
Eta0, line 12: Is TensorFlow used in this example, or is this a relic of some previous revision?

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
```diff
@@ -0,0 +1,68 @@
...
+
+class Transformer(object):
+    def __init__(self):
+        
+        device = "cuda"
+        start = time.time()
+        self.model = GPTJForCausalLM.from_pretrained("/mnt/pvc", torch_dtype=torch.float16).to(device)
```
Eta0, line 19: Double-checking — should this specify a particular folder under `/mnt/pvc`?

online-inference/tensorizer-isvc/gptj_6b_tensorizer/kserve/tensorizer_kserve.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # This ensures that the model is not initialized.
+        with no_init_or_tensor():
+            self.model = AutoModelForCausalLM.from_config(config)
+
+        before_mem = get_mem_usage()
+
+        # Lazy load the tensors from S3 into the model.
+        start = time.time()
+        deserializer = TensorDeserializer("/mnt/pvc/gptj.tensors", plaid_mode=True)
```
Eta0, line 44: Here and in the Flask example the comment says it's loading from S3, and then proceeds to load from a PVC. Which one was intended?

Eta0 (changes requested):

online-inference/tensorizer-isvc/gptj_6b_tensorizer/flask/transformer_flask.py
Eta0, line 1: This file is misnamed, it should be `tensorizer_flask.py`

Eta0 (changes requested):

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
Eta0: I guess a further question is: at what step would these be downloaded into the PVC, and where is that PVC defined? Maybe I missed something.

online-inference/tensorizer-isvc/gptj_6b_hf/kserve/hf-pvc.yml
Eta0, line 1: The names `hf-pvc.yml` and `tensorizer-pvc.yml` sound like manifests for PVCs, but are instead `InferenceService` definitions; this might be misleading.

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
rtalaricw: Yes, the model was downloaded and stored using `save_pretrainined` in the /mnt/pvc path. Once we scope out docs and how to make this customer facing, we can make it a specific path

online-inference/tensorizer-isvc/gptj_6b_tensorizer/kserve/tensorizer_kserve.py
rtalaricw: PVC

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
```diff
@@ -0,0 +1,65 @@
...
+import torch
+from flask import Flask
+from transformers import AutoTokenizer, GPTJForCausalLM
+
+class Transformer(object):
+    def __init__(self):
+        
+        device = "cuda"
+        start = time.time()
+        self.model = GPTJForCausalLM.from_pretrained("/mnt/pvc", torch_dtype=torch.float16).to(device)
```
NavarrePratt, line 16: We can save on a bunch of duplicated code by combining the tensorizer and hf inference service scripts.

To do this, we can add a `--tensorizer` command line flag which changes how the model is loaded. Similar to what we do with our stable diffusion inference example: https://github.com/coreweave/kubernetes-cloud/blob/master/online-inference/stable-diffusion/service/service.py

NavarrePratt (changes requested):
Left various nit-pick comments that apply to more than just the line I commented on, but I didn't want to litter the review with similar comments.

Larger comments are around including more code so the whole benchmark can be run off the shelf, and combining the inference services by adding a `--tensorizer` flag.

online-inference/tensorizer-isvc/benchmark/load_test_async.py
```diff
@@ -0,0 +1,65 @@
+import time
+import asyncio
+import random
+import requests
+import aiohttp
+from argparse import ArgumentParser
+
+def timed(func):
```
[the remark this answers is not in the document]
Eta0: Formatting is fixed in #214

online-inference/tensorizer-isvc/benchmark/load_test_async.py
```diff
@@ -0,0 +1,65 @@
...
+            async def fetch(url):
+                async with session.get(url) as response:
+                    print(response.status)
+                    return await response.text()
+            return await asyncio.gather(*[
+                fetch(url) for url in urls
+            ])
+        
+    # call get_all as a sync function to be used in a sync context
+    return (get_all)(urls)
```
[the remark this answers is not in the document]
Eta0: I simplified this function a bit more in #214 since it doesn't need a nested definition, so this is fixed there too.

online-inference/tensorizer-isvc/gptj_6b_hf/kserve/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ghcr.io/coreweave/ml-containers/torch:bb02bee-base-cuda11.8.0-torch2.0.0-vision0.15.1-audio2.0.1
+
+RUN apt update && apt install -y git curl
```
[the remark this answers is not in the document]
rtalaricw: We need `curl` as a requirement for tensorizer. I can remove `git`

online-inference/tensorizer-isvc/gptj_6b_hf/kserve/hf-isvc.yml
```diff
@@ -0,0 +1,41 @@
+apiVersion: serving.kubeflow.org/v1beta1
+kind: InferenceService
+metadata:
+  labels:
+    qos.coreweave.cloud/latency: low
```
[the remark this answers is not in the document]
rtalaricw: I think it enables lower latency.

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
rtalaricw: I wanted to keep them separate for usability with the trade off of using some duplicated code.

online-inference/tensorizer-isvc/gptj_6b_hf/kserve/Dockerfile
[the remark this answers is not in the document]
Eta0: `curl` isn't a requirement for tensorizer unless you're doing streaming loads, so if this only loads via PVC then that can be removed too.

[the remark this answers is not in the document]
wbrown: @Eta0 I'd say that is a bit of an over-optimization.

online-inference/tensorizer-isvc/gptj_6b_hf/flask/transformer_flask.py
wbrown: @rtalaricw Not a fan of duplicated code. It leads to maintenance issues.

rtalaricw: I am keeping it separate since we use Flask and KServe. Also we only want the tensorizer example, why would we want HF in Inference Examples? I can just remove HF if we are pushing this to be customer facing.

rtalaricw: Internally, we wanted to compare HF v/s Tensorizer but if we want to add a README and publish to GitBook, we should keep it clean with only tensorizer.

online-inference/tensorizer-isvc/gptj_6b_tensorizer/kserve/requirements.txt
```diff
@@ -0,0 +1,2 @@
+kserve==0.10.1
+transformers==4.27.1tensorizer==1.1.0
```
Eta0, line 2: Typo:
```suggestion
transformers==4.27.1
tensorizer==1.1.0
```

Eta0 (changes requested):

Eta0:
Quick note: I disabled `debug` mode on the Flask servers in f2855e0496361c2a502c1b99cb067fecfdbcbde5 because it allows arbitrary code execution. Snyk is appeased with the code here now.

Also, fixed a bug over in 0eca5d326b0582c827ab989f5b589ecaa87eebfe where `tensorizer_kserve.py` couldn't predict because the function signature for `predict` was wrong.

rtalaricw:
Addressed all comments, removed readiness/liveness since we do not want to wait for a set period of time before serving inference. I updated the images and left HF for now. I think we should ideally remove it so we keep it related to tensorizer since we already have HF inference examples.

online-inference/tensorizer-isvc/benchmark/load_test.py
```diff
@@ -0,0 +1,203 @@
...
+                    )
+            except requests.ConnectTimeout:
+                logger.warning(f"Request to {url} timed out")
+
+    return times
+
+
+def random_inference_url(base_url: str) -> str:
+    query = urllib.parse.quote(random.choice(inputs))
+    return f"{base_url}/predict/{query}"
```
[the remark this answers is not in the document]
rtalaricw: It only works for Flask right now.

online-inference/tensorizer-isvc/model-download/model-download-job.yaml
```diff
@@ -0,0 +1,43 @@
...
+            mountPath: /mnt
+        resources:
+          requests:
+            cpu: 8
+            memory: 64Gi 
+            nvidia.com/gpu: 1
+          limits:
+            cpu: 8
+            memory: 64Gi 
+            nvidia.com/gpu: 1
```
Eta0, line 23: These resources are probably overkill for the model downloader. I don't think it should need a GPU at all.

online-inference/tensorizer-isvc/tensorizer_hf_isvc/load_model.py
```diff
@@ -0,0 +1,60 @@
...
+        model = GPTJForCausalLM.from_pretrained(
+            "/mnt/pvc", torch_dtype=torch.float16
+        ).to(DEVICE)
+        print(f"Start time for model load from HF: {time.time() - start} seconds")
+
+        return model
+
+    else:
+        model_ref = "EleutherAI/gpt-j-6B"
+        config = AutoConfig.from_pretrained(model_ref)
```
NavarrePratt, line 34: nit: you don't need the `else` here bc you're returning in the `if`, so can save on another level of indentation.

online-inference/tensorizer-isvc/tensorizer_hf_isvc/load_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+"""
+Loads the model using Tensorizer or Huggingface.
+
+params:
+- model_load_type (str): Type of model load [Options: "tensorizer", "hf"]
+"""
+
+
+def load_model_based_on_type(model_load_type: str = "tensorizer"):
```
NavarrePratt, line 20: This function sits on its own, so can we add a `path` argument that we can default to `/mnt/pvc`?

online-inference/tensorizer-isvc/tensorizer_hf_isvc/kserve/kserve_api.py
```diff
@@ -0,0 +1,67 @@
...
+    def load(self):
+        logger.info(f"Loading {MODEL_NAME}")
+
+        self.model = load_model_based_on_type(model_load_type=MODEL_LOAD_TYPE)
+
+        self.model.eval()
+        torch.manual_seed(100)
+
+        self.tokenizer = AutoTokenizer.from_pretrained("/mnt/pvc/tokenizer")
+        self.eos = self.tokenizer.eos_token_id
```
NavarrePratt, line 33: nit: these instance variables should be initialized in the `init` function.

NavarrePratt (commented):
lgtm, just a few minor nit-picks.

NavarrePratt (approved):

online-inference/tensorizer-isvc/tensorizer_hf_isvc/kserve/kserve_api.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.ready = True
+
+    def predict(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
+        if "text" in payload:
```
harubaru, line 41: Might want to assert that `payload` is actually a `Dict` to avoid type errors just in case.

harubaru (changes requested):
Pretty good changes, other than the feedback I left it looks good.

harubaru (approved):

Eta0:
I added some fixes so that:
- The `kserve_api` code conforms to [KServe's V1 inference protocol](https://kserve.github.io/website/0.10/modelserving/data_plane/v1_protocol/)
  - I.e. input is in the form `{"instances": [...]}` and output is in the form `{"predictions": [...]}`
  - This also means it can now process requests in bulk
- Validation of the KServe payload as a `dict` is done accurately using `isinstance(payload, Dict)`
  - `type(payload) == Dict` is always `False` because `Dict` is a generic and not a real instantiable type, but it is nonetheless runtime-checkable via `isinstance`
- The Flask server is no longer in `debug` mode for the same reason as the change [here](https://github.com/coreweave/kubernetes-cloud/pull/204#issuecomment-1570960056)
- `load_model.py` is refactored a bit to use Google's docstring format and to check for errors without using `assert`
  - Raising a `ValueError` is a bit more descriptive, and `assert` statements can technically be disabled at runtime depending on Python's flags

Eta0:
Added some more fixes and improvements, woo. There's still more I'd like to comment on beyond these changes, so I'll put up the rest of my review once I have time.

## Fixes

- 99f17344b086b9b773094934545f058547aa8815 Console logs now use unbuffered output.
- 93e36ee1b799aed2db0b5b347e5e5acb549ba951 `kserve_api` now uses the correct path to the downloaded tokenizer files.
- 9ca08f96eb04f1bc9671443ec27db7e530f4f053 The model downloader job now requests enough RAM to work.
  - It needs about ~35 GiB, and requests 40 GiB.
- 150584af0265b9865e11c0dc600b762eb0ca83d5 The benchmark code can now find `inputs.txt` regardless of the working directory.
- 671f76a27ade4d2c733d75ac42bb8e6ff5e8c166 `benchmark/load_test.py` now treats `--sync` and `--async` as mutually exclusive arguments, and defaults `--requests` to 100.
  - Previously, the last option specified of either `--sync` or `--async` took priority, and `--requests` had no default.

## Improvements

- 3ff8e9a187b5bb44d9ddd1aa72cfb6af19695d5f The Dockerfiles are now organized for better caching during builds.
  - Dependency installations are cached so that they only need to change if the `requirements.txt` files change.
- 05973d182afaf738fe4a5ff39b409006c2952b28 The model config file in the PVC from the model downloader is now reused for loading tensorized models, rather than downloading it again by model ref at runtime.
- ea6bc48f3896402ca1037e812ebbde4469c10697 KServe-based `InferenceService`s may now be benchmarked using `benchmark/load_test.py` in addition to Flask-based ones.
  - Either `--flask` or `--kserve` must be specified along with `--url` when launching the benchmark to determine how to format the requests.
  - KServe is now the default `InferenceService` demonstrated in the README.

wbrown:
@Eta0 Are we good to merge this past the `black -l 80` reformatting?

online-inference/tensorizer-isvc/tensorizer_hf_isvc/kserve/kserve_api.py
```diff
@@ -0,0 +1,78 @@
...
+        self.ready = True
+
+    def validate(self, payload: Dict):
+        # Ensure that the request has the appropriate type to process
+        if not isinstance(payload, Dict):
+            raise kserve.errors.InvalidInput("Expected payload to be a dict")
+        return super().validate(payload)
+
+    def predict(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
+        inputs = payload.get("instances") or ["Please input some text"]
```
Eta0, line 48: It was unexpected to me that the default for this file is to perform inference on the string `"Please input some text"` when no input is provided. There's nothing wrong with it; it's nice to be able to test the endpoint even if you aren't quite sure how to pass input in the right format for it, but I was just wondering if that was what the original desired behaviour was, versus, like, returning an error code or the string `"Please input some text"` without performing inference.

(Note: I refactored this particular line of code, so I'm the `git blame` "author" for it, but it was just preserving the original functionality in this file).

Eta0 (approved):
Left a few comments on parts that still catch my eye but don't need changes. Looks good to merge as soon as the Docker image tags are updated in `model-download/model-download-job.yaml` and the four `*-isvc.yaml` files under `tensorizer_hf_isvc` to use the new code. Hooray!

online-inference/tensorizer-isvc/README.md
```diff
@@ -0,0 +1,48 @@
...
+  - `kubectl apply -f pvc.yaml`
+- Download the model to the PVC
+  - `kubectl apply -f model-download/model-download-job.yaml`
+- Run the HuggingFace InferenceService (currently using KServe)
+  - `kubectl apply -f tensorizer_hf_isvc/kserve/hf-isvc.yaml`
+- Or, run the Tensorizer InferenceService (currently using KServe)
+  - `kubectl apply -f tensorizer_hf_isvc/kserve/tensorizer-isvc.yaml`
+- View the InferenceService deployment information and URL
+  - `kubectl get isvc`
+  - `http://` may be required in place of `https://` when connecting to the displayed URL
```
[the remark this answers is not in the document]
rtalaricw: I think both work since the Inference Service setups a certificate for HTTPS. I will change the docs to reflect either/or

PR #205, opened 2023-05-26
Finetuner inference tweaks

Eta0:
Implementing the fixes and tweaks I suggested for finetuner inference in #203.

- Ongoing inference no longer blocks the HTTP server's event loop
- Command line arguments instead of environment variables
  - Extra: IP address can now be specified as a command line argument
  - Extra: Inference on CPU (device ID -1) is fixed by not casting the model to fp16 on CPU
- ~~Renamed the container from `kfserving-container` to `http-inference-container`~~
  - Reverted for compatibility with the Kubernetes-level `kfserving` integration

### DIFF
--- a/finetuner-workflow/finetune-workflow.yaml
+++ b/finetuner-workflow/finetune-workflow.yaml
@@ -573,7 +573,7 @@ spec:
                           values:
                             - "{{workflow.parameters.region}}"
             containers:
-              - name: http-inference-container
+              - name: kfserving-container
                 image: "{{workflow.parameters.finetuner_image}}:{{workflow.parameters.finetuner_tag}}"
                 imagePullPolicy: IfNotPresent
                 command: ["/usr/bin/bash", "-c"]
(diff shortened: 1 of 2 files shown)


finetuner-workflow/finetune-workflow.yaml
```diff
@@ -573,18 +573,16 @@ spec:
...
                 imagePullPolicy: IfNotPresent
                 command: ["/usr/bin/bash", "-c"]
-                args: ["/usr/bin/python3 /app/inference.py"]
+                args: ["/usr/bin/python3 /app/inference.py --model=${INFERENCE_MODEL} --port=${INFERENCE_PORT}"]
                 env:
                   - name: INFERENCE_PORT
                     value: "80"
                   - name: INFERENCE_MODEL
                     value: /mnt/pvc/{{ inputs.parameters.model_path }}
-                  - name: STORAGE_URI # KFServing mounts the PVC at /mnt/pvc/
```
harubaru, line 586: This env var needs to be kept as it allows for mounting the PVC so we can load models from it.

finetuner-workflow/finetuner/inference.py
```diff
@@ -34,25 +65,21 @@ class Completion(BaseModel):
...
-    "port": int(os.getenv("INFERENCE_PORT", 80))
-}
-
 model = pipeline(
     "text-generation",
-    model=args["model"],
-    torch_dtype=torch.float16,
-    device=args["device"]
+    model=args.model,
+    torch_dtype=None if args.device_id == -1 else torch.float16,
```
harubaru, line 71: Nice catch!

harubaru (changes requested):

harubaru (approved):

PR #215, opened 2023-06-01
perf(tensorizer-isvc): Use `gunicorn` as a WSGI server with Flask

Eta0:
## WSGI Server Performance
Proposed change to the Flask server setups in #204 for `tensorizer-isvc`.

Production deployments usually run Flask through `gunicorn` rather than Flask's embedded webserver for better performance. Setup for it is fairly trivial (adding `gunicorn` as a dependency and launching via `gunicorn` or `python -m gunicorn`), so since the deployed server in this example is expected to be benchmarked, this PR simply switches the Flask `InferenceService`s' Dockerfiles to launch through `gunicorn` instead.

### DIFF
--- a/online-inference/tensorizer-isvc/gptj_6b_hf/flask/Dockerfile
+++ b/online-inference/tensorizer-isvc/gptj_6b_hf/flask/Dockerfile
@@ -9,4 +9,4 @@ RUN pip install --no-cache-dir --upgrade pip && \
 
 COPY transformer_flask.py ./
 
-ENTRYPOINT ["python", "-m", "gunicorn", "-w8", "-b0.0.0.0", "transformer_flask:app"]
+ENTRYPOINT ["python", "-m", "gunicorn", "-w1", "-b0.0.0.0", "transformer_flask:app"]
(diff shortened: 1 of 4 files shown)


Eta0:
The reason this is a PR rather than a push straight onto #204 is because I'm not sure this is necessary or if it works well in this particular case. I haven't benchmarked it yet (ironically), and the configuration might need tweaking since large model files are likely not viable to load in multiple processes concurrently, which is `gunicorn`'s standard model for concurrency (instead, `gunicorn` is currently set to use a single worker in this PR, which was Flask's embedded server's default concurrency last time I checked).

TL;DR feel free to close this instead of merging if it doesn't seem worth the time.

rtalaricw:
> The reason this is a PR rather than a push straight onto #204 is because I'm not sure this is necessary or if it works well in this particular case. I haven't benchmarked it yet (ironically), and the configuration might need tweaking since large model files are likely not viable to load in multiple processes concurrently, which is `gunicorn`'s standard model for concurrency (instead, `gunicorn` is currently set to use a single worker in this PR, which was Flask's embedded server's default concurrency last time I checked).
> 
> TL;DR feel free to close this instead of merging if it doesn't seem worth the time.

@Eta0 This is perfect. I was literally going to do this. Thanks. I will test it out by EOW. Got a lot of other stuff too.

rtalaricw (approved):

PR #217, opened 2023-06-06
feat(finetuner): add adjustable warmup

harubaru:
This PR adds a new argument to #128 which allows clients to adjust the ratio of warm up steps to training steps which is fairly important since the GPT finetuner uses an adaptive optimizer (Adam), plus this also addresses @Eta0's feedback on a duplicate workflow argument.

### DIFF
--- a/finetuner-workflow/finetune-workflow.yaml
+++ b/finetuner-workflow/finetune-workflow.yaml
@@ -40,9 +40,6 @@ spec:
     # and `pile`. It defaults to the model's tokenizer data if not provided.
     - name: tokenizer
       value: ''
-    # Save model checkpoint every N steps.
-    - name: save_steps
-      value: '500'
     # How should the tokenizer order the dataset inputs? Options available are:
     # `size_ascending`, `size_descending`, `name_ascending`, `name_descending`,
     # `random`, and `none`. `shuffle` implements `none` while also assigning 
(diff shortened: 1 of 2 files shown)


finetuner-workflow/finetune-workflow.yaml
```diff
@@ -275,6 +275,7 @@ spec:
               --train-ratio=1.0
               --eot='{{workflow.parameters.eot_token}}'
               --pad='{{workflow.parameters.pad_token}}'
+              --warmup_ratio={{workflow.parameters.warmup_ratio}}
```
Eta0, line 278:
```suggestion
              --warmup-ratio={{workflow.parameters.warmup_ratio}}
```

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -988,7 +999,7 @@ def evaluate(
     gradient_checkpointing=True,
 
     # Learning rate scheduler arguments
-    warmup_steps=8,
+    warmup_steps=warmup_steps,
```
Eta0, line 1002: `TrainingArguments` supports specifying the `warmup_ratio` directly as a float:
```suggestion
    warmup_ratio=float(args.warmup_ratio),
```

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -972,6 +979,10 @@ def evaluate(
         )
     )
 
+# Calculate how many warmup steps we need to do using the warmup ratio.
+training_steps = int((args.epochs * len(train_dataset)) / (args.gradients * bs * world_size))
+warmup_steps = int(args.warmup_ratio * training_steps)
+
```
Eta0, line 985: This calculation can be skipped by specifying `warmup_ratio` directly in the `TrainingArguments` constructor (see other comment)
```suggestion
```

Eta0 (changes requested):
Left some comments on how it can be simplified a bit, but besides that, it looks like a good change!

finetuner-workflow/finetuner/finetuner.py
harubaru: Oh wow didn't know that

Eta0 (approved):

wbrown (approved):
Good. :+1:

PR #222, opened 2023-06-09
Update CUDA SSH Example

Eta0:
# CUDA SSH

This PR brings the deployment for the [`cuda-ssh` docs example](https://docs.coreweave.com/coreweave-kubernetes/examples/cuda-ssh) up to date.

## New Container

The CUDA SSH container has been re-homed from:
- [docker.io/atlanticcrypto/cuda-ssh-server](https://hub.docker.com/r/atlanticcrypto/cuda-ssh-server)

To:

- [ghcr.io/coreweave/ml-containers/cuda-ssh](https://github.com/coreweave/ml-containers/pkgs/container/ml-containers%2Fcuda-ssh)

With the new container based on the NCCL version of [ghcr.io/coreweave/ml-containers/torch](https://github.com/coreweave/ml-containers/pkgs/container/ml-containers%2Ftorch). There is an alternative version of the container built off of the [torch image](https://github.com/coreweave/ml-containers/pkgs/container/ml-containers%2Ftorch)'s base edition that could also be used. The corresponding [coreweave/ml-containers PR is located here](https://github.com/coreweave/ml-containers/pull/20).

## Security Updates

This also includes security updates. Private host keys are no longer embedded in the container image, and instead are generated during initialization. Additionally, public key authentication is now used instead of a randomly-generated password.

To support switching to public key authentication, the accompanying docs for this example will need to be updated as well.

### Documentation Details

The new series of bash commands to set up the server is:

```bash
# Create persistent storage volumes
kubectl apply -f sshd-root-pvc.yaml -f sshd-data-pvc.yaml

# Create the SSH service and public IP
kubectl apply -f sshd-service.yaml

# Launch the SSH server
kubectl apply -f sshd-deployment.yaml

# Wait for the SSH server to come up
kubectl rollout status deployment/sshd

# Add your SSH public key
kubectl exec -i deployment/sshd -c sshd -- /bin/tee --append /root/.ssh/authorized_keys < ~/.ssh/id_rsa.pub

# Print the public IP address
kubectl get service/sshd -o 'jsonpath=service/sshd public IP: {.status.loadBalancer.ingress[0].ip}{"\n"}'

# Print the host key fingerprints
kubectl exec deployment/sshd -c sshd -- find /etc/ssh -type f -name 'ssh_host_*_key.pub' -exec ssh-keygen -lf '{}' ';'

# Connect
ssh root@[public IP]
```

Notes:

- The "Add your SSH public key" step requires the user to have a public SSH key at `~/.ssh/id_rsa.pub`. It's likely that they already have one if they're trying to use an SSH server, but if they don't, running `ssh-keygen` will create one at that path.
  - That line can be modified to use any path to an SSH public key.
- An alternate version of the "Print the public IP address" step could also be `kubectl get service/sshd`, which is simpler and gives more information, whereas the one shown above gives exactly the relevant IP and nothing else.
- The commands listed are for bash, but use minimal shell features.
  - If Windows equivalents are needed, the exact same commands will work in `powershell` with one modification for PowerShell's way of doing input redirection:
    ```diff
    - kubectl exec -i deployment/sshd -c sshd -- /bin/tee --append /root/.ssh/authorized_keys < ~/.ssh/id_rsa.pub
    + Get-Content -Encoding utf8 ~/.ssh/id_rsa.pub | kubectl exec -i deployment/sshd -c sshd -- /bin/tee --append /root/.ssh/authorized_keys
    ```

## Additional Enhancements

- The container now runs with an init process,
- The node affinity has been changed to require `RTX_A5000` GPUs,
- The region selector label has been updated, and
- The pod requests a more appropriate 2.5–7 CPUs and 64–128 GiB RAM to accompany its 6 GPUs, rather than 1.5–7 CPUs and 4–8 GiB RAM.

### DIFF
--- a/cuda-ssh/sshd-data-pvc.yaml
+++ b/cuda-ssh/sshd-data-pvc.yaml
@@ -3,7 +3,7 @@ kind: PersistentVolumeClaim
 metadata:
   name: sshd-data-pv-claim
 spec:
-  # https://docs.coreweave.com/coreweave-kubernetes/storage
+  # https://docs.coreweave.com/storage/storage
   storageClassName: block-nvme-ord1
   accessModes:
     - ReadWriteOnce
--- a/cuda-ssh/sshd-data-pvc.yaml
+++ b/cuda-ssh/sshd-data-pvc.yaml
@@ -4,7 +4,7 @@ metadata:
   name: sshd-data-pv-claim
 spec:
   # https://docs.coreweave.com/storage/storage
-  storageClassName: block-nvme-ord1
+  storageClassName: block-hdd-ord1
   accessModes:
     - ReadWriteOnce
   resources:
--- a/cuda-ssh/sshd-deployment.yaml
+++ b/cuda-ssh/sshd-deployment.yaml
@@ -17,7 +17,7 @@ spec:
       terminationGracePeriodSeconds: 10
       initContainers:
       - name: init
-        image: ghcr.io/coreweave/ml-containers/cuda-ssh:es-cuda-ssh-87f59b9
+        image: ghcr.io/coreweave/ml-containers/cuda-ssh:es-cuda-ssh-8b9c879-torch-ceeb8c2-nccl-cuda11.8.0-nccl2.16.2-1-torch2.0.1-vision0.15.2-audio2.0.2
         command: ["/bin/bash", "-c"]
         args:
           - |
@@ -40,7 +40,7 @@ spec:
         command: ["/usr/bin/tini", "--"]
         args: ["service", "ssh", "start", "-D"]
         tty: true
-        image: ghcr.io/coreweave/ml-containers/cuda-ssh:es-cuda-ssh-87f59b9
+        image: ghcr.io/coreweave/ml-containers/cuda-ssh:es-cuda-ssh-8b9c879-torch-ceeb8c2-nccl-cuda11.8.0-nccl2.16.2-1-torch2.0.1-vision0.15.2-audio2.0.2
         ports:
           - name: sshd
             containerPort: 22
--- a/cuda-ssh/sshd-deployment.yaml
+++ b/cuda-ssh/sshd-deployment.yaml
@@ -17,7 +17,7 @@ spec:
       terminationGracePeriodSeconds: 10
       initContainers:
       - name: init
-        image: ghcr.io/coreweave/ml-containers/cuda-ssh:es-cuda-ssh-8b9c879-torch-ceeb8c2-nccl-cuda11.8.0-nccl2.16.2-1-torch2.0.1-vision0.15.2-audio2.0.2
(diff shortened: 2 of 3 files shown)


Eta0:
This PR is a draft because it is not ready to be merged until the documentation can be updated. If @coreweave/technical-writers could give this a look, that'd be swell!

microcosem:
On it, thanks Eta! :) I'll get this done today.

microcosem:
Edits submitted by reviewer. Second draft underway.

rtalaricw (approved):
LGTM

Eta0:
> Here's a quandary for the other reviewers when they have time to look:
> 
> For the last 2 years, the documentation has been inaccurate, and has stated that `sshd-data-pvc.yaml` allocates HDD storage, when it's actually been [NVMe storage since this commit](https://github.com/coreweave/kubernetes-cloud/commit/904b0c1f01b7389a755d4e462ec69939d27df6c7), which may have been an accident itself.
> 
> I'm not sure whether the documentation or the code should be the one to change. I think it would be a good example of "responsibly" mixing SSD and HDD storage volumes if the code were to actually do that. At the same time, a big chunk of extra SSD storage is fairly nice for a development machine (though, of course, also more expensive).
> 
> I'm leaning towards just switching the manifest back to use an HDD. Thoughts?

Switched it back to HDD storage in 15e08673c1cb8828551fb9bed1e3f5dbb85cb870.

rtalaricw:
> > Here's a quandary for the other reviewers when they have time to look:
> > For the last 2 years, the documentation has been inaccurate, and has stated that `sshd-data-pvc.yaml` allocates HDD storage, when it's actually been [NVMe storage since this commit](https://github.com/coreweave/kubernetes-cloud/commit/904b0c1f01b7389a755d4e462ec69939d27df6c7), which may have been an accident itself.
> > I'm not sure whether the documentation or the code should be the one to change. I think it would be a good example of "responsibly" mixing SSD and HDD storage volumes if the code were to actually do that. At the same time, a big chunk of extra SSD storage is fairly nice for a development machine (though, of course, also more expensive).
> > I'm leaning towards just switching the manifest back to use an HDD. Thoughts?
> 
> Switched it back to HDD storage in [15e0867](https://github.com/coreweave/kubernetes-cloud/commit/15e08673c1cb8828551fb9bed1e3f5dbb85cb870).

Sorry, I meant to reply to your comment. HDD is fine. We can mention this in the docs and provide an option to choose between HDD or NVME probably as a patch or whatever you think works

microcosem:
@Eta0 as per @rtalaricw 's note, I've added a note that suggests you can define the volume as an NVME volume, but left everything else referencing HDD in tact since you've fixed it. :) If this is all good, I'll go ahead and merge everything!

Eta0:
@microcosem I just updated the manifest snippets in the docs CR to reflect the changes as well, and it should be good to merge now ✨

microcosem:
Merging!

PR #232, opened 2023-07-10
feat(downloader): update model downloader in bloom example for parall…

jfeldman325:
Replaces sequential download with 🦾 and 💨 s5cmd which will handle the downloads in parallel. Produces significant speed ups :)


online-inference/bloom-176b/model/scripts/pytorch_model_00072-of-00072.bin
harubaru, line 1: I think this was accidentally left in

harubaru (changes requested):

PR #236, opened 2023-07-24
Remove docs because GitBook is unlinked.

dfinster:
This PR removes the docs directory. 

The Technical Writing team has taken advantage of the new GitBook monorepo feature, which allows us to sync multiple spaces to a single GitHub repo. We have many internal-only spaces, so we've created a new private repo, [gitbook-sync](https://github.com/coreweave/gitbook-sync), dedicated to this purpose for security. The gitbook-sync repo is dedicated as a backup of all GitBook spaces.

The GitBook monorepo is now established and GitBook no longer syncs to [kubernetes-cloud/docs](https://github.com/coreweave/kubernetes-cloud/tree/master/docs). 

The docs folder is no longer in sync with the CoreWeave documentation, and should be deleted.


Eta0 (changes requested):
There are other files and directories left over that are or were associated with GitBook, but are outside of the `docs/` directory. These are a few I notice:
- `.gitbook.yaml`
- `.gitbook/`
- `coreweave-kubernetes/`
- `resources/`
- `virtual-servers/`
- Part of `online-inference/custom-basnet/`, mixed in with the example code somehow

You may wish to double-check that those all migrated over properly (or are no longer needed), and then remove them here.

Eta0:
There is also the `argo-workflow` directory, which seems to correspond to [docs.coreweave.com/cloud-tools/argo](https://docs.coreweave.com/cloud-tools/argo), but the documentation page does not link back to `kubernetes-cloud` at all—instead using inline code blocks for the one relevant YAML manifest there. That directory could probably be removed from `kubernetes-cloud` too, assuming it is no longer needed.

dfinster:
> There are other files and directories left over that are or were associated with GitBook, but are outside of the `docs/` directory.

I reviewed that list and agree. I pushed another commit to this PR that removes those. I only removed a portion of  `online-inference/custom-basnet/` that I could verify was GitBook related. 

It looks like these scattered directories are an artifact of how GitBook has evolved at CoreWeave. The settings in GitBook clearly show that all content should have been contained to the `/docs` folder, but this obviously isn't true because some of the content you've identified is recent. I think it's the result of previous migrations and setting changes. 

Anything else GitBook related in `kubernetes-cloud` can be deleted. GitBook is currently set to only push to GitHub, it never pulls. We are treating the `gitbook-sync` repo as a live backup, and the main source of truth is stored in GitBook. No migration from this repo was actually required, we simply changed the target to the new repo, and GitBook made a fresh commit with the current state. We have all the history and CR branches in GitBook and do not need them here.

dfinster:
> There is also the `argo-workflow` directory, which seems to correspond to [docs.coreweave.com/cloud-tools/argo](https://docs.coreweave.com/cloud-tools/argo), but the documentation page does not link back to `kubernetes-cloud` at all—instead using inline code blocks for the one relevant YAML manifest there. That directory could probably be removed from `kubernetes-cloud` too, assuming it is no longer needed.

I have mixed feelings about this. I think there's some value in keeping `gpu-say-workflow.yaml` here and we should update the GitBook docs to link to it. In particular, line endings can get weird when copy-pasting from HTML sometimes, so there's good value in having both. Let's get @microcosem to offer an opinion before I delete this one.

microcosem (approved):

microcosem:
> > There is also the `argo-workflow` directory, which seems to correspond to [docs.coreweave.com/cloud-tools/argo](https://docs.coreweave.com/cloud-tools/argo), but the documentation page does not link back to `kubernetes-cloud` at all—instead using inline code blocks for the one relevant YAML manifest there. That directory could probably be removed from `kubernetes-cloud` too, assuming it is no longer needed.
> 
> I have mixed feelings about this. I think there's some value in keeping `gpu-say-workflow.yaml` here and we should update the GitBook docs to link to it. In particular, line endings can get weird when copy-pasting from HTML sometimes, so there's good value in having both. Let's get @microcosem to offer an opinion before I delete this one.

Yeah - this is _probably_ the better solution for now since all other example code lives in kubernetes-cloud, so this should, too. We can link out to it if it requires better completion.

For now, let's just leave that there?

dfinster:
@Eta0 This is ready for another review based on @microcosem comments.

Eta0 (approved):
Looks good 👍

PR #248, opened 2023-08-28
(fix) changing base image to ubuntu and addressing minor bugs

anguyencoreweave:
Slack Thread: https://coreweave.slack.com/archives/C03L6UD9EJ1/p1693229646897549

This PR intends to unblock users doing the stable diffusion tutorial who want to build their own image of model-downloader locally and push to a public repo like Docker Hub. Users following the current instructions will be blocked due to a bug where the requirements.txt file is not being copied over properly into the working directory AND an incompatibility between alpine and pytorch/numpy. 

The issue is addressed here by modifying the base image to ubuntu, and copying over the proper text files into the working directory to specify requirements.

I was able to build and push an image of `model-downloader` successfully using my modifications.

### DIFF
--- a/online-inference/stable-diffusion/Dockerfile.downloader
+++ b/online-inference/stable-diffusion/Dockerfile.downloader
@@ -1,7 +1,11 @@
-FROM python:3.9.13-alpine3.16
+FROM ubuntu
+RUN apt-get update 
+RUN apt-get install -y python3 python3-pip 
 RUN mkdir /app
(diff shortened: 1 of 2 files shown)


Oogy:
@anguyencoreweave Can we close this? Seems the `Dockerfile.downloader` image is no longer used.

anguyencoreweave:
Cancelling PR.

PR #303, opened 2023-12-20
HF LLM Overhaul

harubaru:
This PR adds a simple HuggingFace LLM inference server to the examples and it also removes the older and broken GPT-2 example. A majority of the code has been ported over from the Stable Diffusion inference service in order to provide a simple easy-to-use example as a way to introduce people to running inference services on CoreWeave.


online-inference/hf-llm/service/service.py
```diff
@@ -0,0 +1,140 @@
...
+            top_k=completion.top_k,
+            repetition_penalty=completion.repetition_penalty,
+            do_sample=completion.do_sample,
+            penalty_alpha=completion.penalty_alpha,
+            num_return_sequences=completion.num_return_sequences,
+            stop_sequence=completion.stop_sequence
+        )
+        return Response(content=json.dumps(output), media_type="application/json", status_code=200)
+    except Exception as e:
+        return Response(content=str(e), media_type="text/plain", status_code=500)
```
Eta0, line 135: You can provide more detail via the `logging` module while also preventing any information from leaking in server responses with something like this:
```suggestion
        import traceback

        logger.error(traceback.format_exc())
        return Response(
            content=(
                f"Server encountered an error of type {type(e).__name__}."
                "\nSee logs for details."
            ),
            media_type="text/plain",
            status_code=500,
        )
```

online-inference/hf-llm/serializer/requirements.txt
```diff
@@ -0,0 +1,3 @@
+torch==2.0.0
```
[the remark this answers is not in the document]
harubaru: I'll set it to `<2.3.0` since it might confuse users if a torch 3.0.0 exists lol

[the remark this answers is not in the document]
wbrown: I've had problems with it pulling down later torch'es if we don't specify exactly or use ~=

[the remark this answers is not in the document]
Eta0: That's a generally a side effect of `pip install -U`/`pip install --upgrade`; upgrading will pull newer versions of `torch` whenever available, but regular `pip install` will keep the existing installed version unless it's incompatible with the other requirements.

online-inference/hf-llm/02-inference-service.yaml
Eta0, line 1: This file needs its references to secrets updated to match the coalesced setup

online-inference/hf-llm/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ghcr.io/coreweave/ml-containers/torch:afecfe9-base-cuda11.8.0-torch2.0.0-vision0.15.1
```
Eta0, line 1: The `ml-containers` Dockerfile currently uses torch-base with CUDA 12.0.1 (we could update these to a newer build too)

```suggestion
FROM ghcr.io/coreweave/ml-containers/torch:afecfe9-base-cuda12.0.1-torch2.0.0-vision0.15.1
```

Eta0 (changes requested):

online-inference/hf-llm/service/service.py
```diff
@@ -0,0 +1,149 @@
...
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+
+@app.get("/")
+def get_health():
+    logger.info("Health check requested")
```
Eta0, line 115: This shouldn't normally be necessary to know (and this event will be happening continuously) so this is fine as a debug log statement
```suggestion
    logger.debug("Health check requested")
```

online-inference/hf-llm/02-inference-service.yaml
```diff
@@ -39,20 +39,20 @@ spec:
...
           valueFrom:
             secretKeyRef:
-              name: s3-secret-key
+              name: s3-credentials
               key: secret_key
               optional: true
         - name: AWS_HOST
           valueFrom:
             secretKeyRef:
               name: s3-host-url
```
Eta0, line 54:
```suggestion
              name: s3-credentials
```

Eta0 (changes requested):
Testing it start-to-finish at the moment, noticed this in the meantime

Eta0 (changes requested):
Added fixes for some issues in #306, please take a look.

Eta0:
Seems all good once #306 is merged and the container image is updated to reflect the changes.
Noticed some weird inference behaviour in always getting the same results from prompts (and always extremely short ones when serving `--model-uri=EleutherAI/pythia-70m`, like the prompt unchanged or with one extra word), but the Kubernetes deployment itself and model loading are fine, so it'd be fine to merge for now and poke at that strangeness afterward.

wbrown:
> Seems all good once #306 is merged and the container image is updated to reflect the changes. Noticed some weird inference behaviour in always getting the same results from prompts (and always extremely short ones when serving `--model-uri=EleutherAI/pythia-70m`, like the prompt unchanged or with one extra word), but the Kubernetes deployment itself and model loading are fine, so it'd be fine to merge for now and poke at that strangeness afterward.
@Eta0 This is often a characteristic of sampling behaviors, especially when combined with very small models.

wbrown:
@Eta0 You need to resolve your requested change if we want to merge this.

Eta0:
> @Eta0 You need to resolve your requested change if we want to merge this.

@wbrown Approve & merge coreweave/ml-containers#52 and I can update the image tag and approve it, that's the last thing that stills needs a change

Eta0 (approved):

Issue #353, opened 2024-02-26
[finetuner-workflow]: issues encountered at download model step


dmarx:
1. given `tensorizer_uri` pointing to the tensorizer s3 bucket and `model` provided as a huggingface identifier, the pipeline incorrectly attempts to use the `model` argument as a path on the PVC.

```
»  k logs llm-finetune-template-fph9l-model-tokenizer-1285859084 -c main
2024/02/26 16:43:36 Tokenizer definition: /finetune-data/models/mistralai/Mistral-7B-v0.1
2024/02/26 16:43:36 Tokenizer input source: /finetune-data/sep23_txt
2024/02/26 16:43:36 Tokenizer output: /finetune-data/sep23_txt-mistralai_Mistral_7B_v0_1-2048-b-1-7c82f97.tokens
2024/02/26 16:43:36 Tokenizer reordering method: 
2024/02/26 16:43:36 Sampling amount (in % tokens kept): 100%
2024/02/26 16:43:36 Resolving /finetune-data/models/mistralai/Mistral-7B-v0.1/config.json... 
2024/02/26 16:43:37 /finetune-data/models/mistralai/Mistral-7B-v0.1/config.json not found, required!
2024/02/26 16:43:37 cannot retrieve required `/finetune-data/models/mistralai/Mistral-7B-v0.1 from config.json`: HTTP status code 404
time="2024-02-26T16:43:37.956Z" level=info msg="sub-process exited" argo=true error="<nil>"
Error: exit status 1
```

2. when `tensorizer_uri` not provided, the model appears to download as it is supposed to from huggingface, but then throws an error that appears to be based on an attempted module name conversion.

```
»  k logs llm-finetune-template-4rxg9-model-downloader-781196140 -c main
2024/02/26 17:22:59 Resolving mistralai/Mistral-7B-v0.1/pytorch_model.bin.index.json... 
2024/02/26 17:22:59 Downloaded mistralai/Mistral-7B-v0.1/pytorch_model.bin.index.json... 24 kB completed.
2024/02/26 17:22:59 Resolving mistralai/Mistral-7B-v0.1/tokenizer.json... 
2024/02/26 17:23:00 Downloaded mistralai/Mistral-7B-v0.1/tokenizer.json... 1.8 MB completed.
2024/02/26 17:23:00 Resolving mistralai/Mistral-7B-v0.1/tokenizer_config.json... 
[...]

PR #376, opened 2024-03-21
LLM Finetuner Cleanup

harubaru:
The primary motivation behind this PR is to not only clean up our LLM Finetuner example by fixing some issues but to also integrate more of Tensorizer's functionality into the finetuner to serve as an example of how Tensorizer can be conveniently used outside of inference usecases.

## Change Overview
- Updated dependencies such as PyTorch, HF Transformers, DeepSpeed, and Accelerate.
- Removed the previous inference service which was originally developed specifically for the LLM Finetuner in favor of our more up to date LLM inference example.
- Addressed https://github.com/coreweave/kubernetes-cloud/issues/353 by redoing the behavior of the `tensorizer_uri` workflow parameter.
- Added support for checkpointing models with Tensorizer.
- The HF LLM inference example now also supports loading models through PVC with a fallback of relying on HF's loading code.
- Integrated vLLM as the inference server which consumes Tensorizer checkpoints.
- Added a maximum batch size argument which caps the batch size estimate which generated a very high estimate for smaller models.
- Support for Mistral models. Training Mistral models is quite slow due to the very large context window which makes the 7B Mistral model require ~96GiB of VRAM to train. In the future, we should be investigating methods to move away from our current form of parallelism.

## Minor TODO
- Checkpoint final artifact to object storage to remove reliance on PVC for inference.
- Serialize trainer state such as LR Scheduler and Optimizer using Tensorizer.
- Documentation regarding the `tensorizer_uri` workflow parameter and other changes.
- Support for the `train_ratio` argument.


finetuner-workflow/finetune-workflow.yaml
```diff
@@ -194,9 +198,13 @@ spec:
     - name: dataset_downloader_tag
       value: 'cd6408a'
     - name: finetuner_image
-      value: 'gooseai/finetuner'
+      value: 'docker.io/harubaru1/finetuner'
```
rtalaricw, line 201: Should we package these images in `ml-containers` instead of using public domains like `docker.io`?

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -194,9 +198,13 @@ spec:
...
     - name: finetuner_tag
-      value: 'b32173f'
+      value: '62'
+    - name: inference_image
+      value: 'docker.io/harubaru1/hf-llm'
```
rtalaricw, line 205: Same here for images being public.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -323,12 +331,13 @@ spec:
...
       parameters:
         - name: model
+        - name: tensorizer_uri
     retryStrategy:
       limit: 1
     container:
       image: "alpine/curl:3.14"
       command: [ "/bin/sh", "-c" ]
-      args: [ "/usr/bin/curl -I -XGET -s -o /dev/null -w \"%{http_code}\" https://accel-object.ord1.coreweave.com/tensorized/{{inputs.parameters.model}}/model.tensors | grep -q 200 && echo -n 'true' > /tmp/output.txt || echo -n 'false' > /tmp/output.txt" ]
+      args: [ "/usr/bin/curl -I -XGET -s -o /dev/null -w \"%{http_code}\" https://accel-object.ord1.coreweave.com/tensorized/{{inputs.parameters.model}}/model.tensors | grep -q 200 || [ -n \"{{inputs.parameters.tensorizer_uri}}\" ] && echo -n 'true' > /tmp/output.txt || echo -n 'false' > /tmp/output.txt" ]
```
rtalaricw, line 340: Do we want to use `accel-object` by default?

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -582,17 +591,19 @@ spec:
...
-                args: ["/usr/bin/python3 /app/inference.py --model=${INFERENCE_MODEL} --port=${INFERENCE_PORT}"]
+                args: ["/usr/bin/python3 /app/service.py --model-uri=${INFERENCE_MODEL} --port=${INFERENCE_PORT} --precision=${PRECISION}"]
                 env:
                   - name: INFERENCE_PORT
                     value: "80"
                   - name: INFERENCE_MODEL
                     value: /mnt/pvc/{{ inputs.parameters.model_path }}
                   - name: STORAGE_URI # KFServing mounts the PVC at /mnt/pvc/
                     value: pvc://{{ workflow.parameters.pvc }}/
+                  - name: PRECISION
```
rtalaricw, line 605: This should be called `INFERENCE_PRECISION`

finetuner-workflow/finetuner/requirements.txt
```diff
@@ -1,10 +1,8 @@
...
 numpy~=1.24.2
 wandb~=0.14.0
 torch==2.0.0
 psutil==5.9.4
-accelerate~=0.17.1
-tensorizer==1.1.0
-fastapi==0.85.1
-uvicorn==0.19.0
+accelerate==0.27.2
+tensorizer==2.8.1
```
rtalaricw, line 7: `tensorizer>=2.8.1`? Does it have to be hard coded to a specific version?

rtalaricw (commented):
Few comments but looks good overall!

Rexwang8:
Nothing stands out to me other than stuff that Rahul mentioned, looks good!

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -1054,6 +1076,7 @@ def collector(data):
 # At the end of it all, record to a `final` output.
 final_path = os.path.join(output_dir, "final")
 trainer.save_model(final_path)
+tensorizer_save(final_path) # Must be invoked manually.
```
sangstar, line 1079: If `save_steps` is a multiple of the number of steps per epoch, unless I'm unaware of how `Trainer` saves checkpoints, it will do a double save at the end, or two quick saves at the end if `num_steps % num_steps_per_epoch` is close to 0. Do we want to possibly skip saving the final checkpoint if this is the case?

sangstar (commented):
LGTM, few questions/nits.

online-inference/hf-llm/service/service.py
```diff
@@ -135,8 +144,10 @@ def load_model(path_uri: str) -> TextGenerationPipeline:
     logger.info(f"Loading {args.precision} model from: {path_uri}")
     if path_uri.startswith("s3://"):
         return load_model_s3(path_uri)
-    else:
+    elif os.path.exists(f"{path_uri}/model.tensors"):
         return load_model_local(path_uri)
+    else:
+        return load_model_hub(path_uri)
```
Eta0, line 150: The `s3://` case accounts for providing a fully specified path such as `s3://tensorized/EleutherAI/pythia-70m/fp16/model.tensors`; this case should do something similar and check if the referred-to path is a file or a directory, e.g.

```suggestion
    if os.path.exists(path_uri):
        if os.path.isfile(path_uri):
            return load_pipeline(path_uri)
        full_path: str = os.path.join(path_uri, "model.tensors")
        if os.path.isfile(full_path):
            return load_pipeline(full_path)
    return load_model_hub(path_uri)
```

This leaves `load_model_hub` as the loading path for `path_uri`s that either:
- Do not exist on the filesystem, or
- Are directories that don't contain a `model.tensors` file

(This series of conditionals could also be written a bit more cleanly with `pathlib`)

online-inference/hf-llm/service/service.py
```diff
@@ -135,8 +144,10 @@ def load_model(path_uri: str) -> TextGenerationPipeline:
     logger.info(f"Loading {args.precision} model from: {path_uri}")
     if path_uri.startswith("s3://"):
         return load_model_s3(path_uri)
-    else:
+    elif os.path.exists(f"{path_uri}/model.tensors"):
         return load_model_local(path_uri)
```
Eta0, line 148: The `load_model_local` case loading a local tensorized model does not cast to or check for the requested precision. (The S3 case handles this by automatically choosing the correct path in the `tensorized` bucket between `.../model.tensors` or `.../fp16/model.tensors`, but for local model loading, that path structure being available is less likely.)

finetuner-workflow/finetuner/requirements.txt
Eta0: `tensorizer>=2.8.1,<3.0.0` should be safe

Eta0 (changes requested):

finetuner-workflow/finetuner/finetuner.py
Eta0: For predictability, there are pros to not skipping it; then something exists at the checkpoint file path and at the `final_path` path, and either can be deleted safely, or otherwise processed programmatically. Though in a storage location where hardlinking is available, that could be used to reduce data duplication while keeping both paths (if it is safe to propagate modifications to either file to either other file).

harubaru: The checkpoint callback is not called if `trainer.save_model()` is invoked manually.

finetuner-workflow/finetune-workflow.yaml
harubaru: Yes, these are just placeholders

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -493,6 +494,26 @@ def compute_loss(self, model, inputs, return_outputs=False):
...
 
+def tensorizer_save(checkpoint_dir: str) -> None:
+    checkpoint_model = trainer.model
+    supported_classes = (PreTrainedModel,)  # TODO: PEFT support.
+    checkpoint_model = checkpoint_model if isinstance(checkpoint_model, supported_classes) else unwrap_model(
+        checkpoint_model)
+
+    serializer = TensorSerializer(f"{checkpoint_dir}/model.tensors")
+    serializer.write_module(checkpoint_model)
+    serializer.close()
```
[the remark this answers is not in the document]
harubaru: Maybe, but that might be out of scope

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -582,31 +702,42 @@ spec:
...
                 ports:
                   - protocol: TCP
                     containerPort: 80
                 livenessProbe:
                   httpGet:
-                    path: /
+                    path: /health
                     port: 80
-                  initialDelaySeconds: 300
+                  initialDelaySeconds: 30
```
rtalaricw, line 734: @harubaru Should we parametrize this in the workflow? I thought 30 seconds was low but curious what you think.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -648,7 +675,7 @@ def __init__(self, path: str, context_length: int = 2048):
         length_mb = os.stat(path).st_size / (1 << 20)
         num_tokens = self.length * context_length
         self._padding_is_ambiguous = tokenizer.pad_token_id == tokenizer.eos_token_id
-        self._pad_token_id = tokenizer.pad_token_id
+        self._pad_token_id = 0xffff  # uint16 max
```
rtalaricw, line 678: We should add a TODO to change this with Llama support especially 3.1

rtalaricw (dismissed):
LGTM. I would love to run this using Argo once merged to `main`

finetuner-workflow/finetune-workflow.yaml
harubaru: It takes about the 30 seconds to load a 7B model, though I could increase it to 60 seconds to add extra legroom.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -217,8 +226,7 @@ spec:
             - name: dest
               value: "/{{workflow.parameters.pvc}}/models/{{workflow.parameters.model}}"
             - name: tokenizer_only
-              value: "{{steps.check-model.outputs.result}}"
```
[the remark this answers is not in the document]
harubaru: Marking as resolved since the `or` has since been replaced: https://github.com/coreweave/kubernetes-cloud/pull/376/commits/d795de4caffb0e6af3f3b925aa4643702996a092

finetuner-workflow/finetuner/finetuner.py
Rexwang8, line 678: this will be uint16 max if the vocab size is below uint16, but will be uint32 max if it is above that limit. if it is exactly uint16, it will be uint32 as well as we would lack the space to inject a pad token in.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -1028,10 +1056,19 @@ def evaluate(
 
 
 def collector(data):
+    input_ids = torch.stack([f[0] for f in data])
+    attention_mask = torch.stack([f[1] for f in data])
+    labels = torch.stack([f[0] for f in data])
+
+    padding_mask = input_ids == 0xffff # TODO: Update to uint32_max when we switch to uint32.
```
Rexwang8, line 1063: we uint32 now

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -891,6 +917,8 @@ def evaluate(
 # use for this model and GPU.
```
Rexwang8, line 917: nit - caps

Rexwang8 (dismissed):
Looks good, add uint32 padding and you're good to go

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -549,11 +616,64 @@ spec:
...
+      image: "{{workflow.parameters.inference_image}}:{{workflow.parameters.inference_tag}}"
+      command: [ "/bin/bash", "-c" ]
+      args:
+        - |
+          cd /app/vllm &&
+          python3 -m examples.tensorize_vllm_model \
+          --model=${MODEL_PATH} \
+          serialize \
+          --serialized-directory /{{workflow.parameters.pvc}} \
+          --suffix vllm
```
dmarx, line 635: `vllm` will already be part of the resulting path: pretty sure the full path created by that script will be `/<serialized-directory>/vllm/<model>/<suffix>`. Suggest using vLLM's recommended convention of `--suffix v1`

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -310,53 +336,90 @@ spec:
...
       limit: 1
     container:
       image: "{{workflow.parameters.model_downloader_image}}:{{workflow.parameters.model_downloader_tag}}"
       command: [ "/ko-app/model_downloader" ]
       args: [ "-model", "{{inputs.parameters.model}}",
               "-dest", "{{inputs.parameters.dest}}",
-              "-tokenizer-only", "{{inputs.parameters.tokenizer_only}}" ]
+              "-tokenizer-only={{inputs.parameters.tokenizer_only}}" ]
+      env:
+      - name: HF_API_TOKEN
```
dmarx, line 413: Getting deja vu here... this is parameter specified by the model_downloader, yeah? If so then this suggestion is out of scope for this PR, but we should align this variable name with the environment variable used by the huggingface_hub, which is HF_TOKEN. see https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables

finetuner-workflow/finetuner/finetuner.py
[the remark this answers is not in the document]
dmarx: agree, let's call that out of scope for now

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -310,53 +336,90 @@ spec:
...
+            - name: model_path
+              value: "/{{workflow.parameters.pvc}}/finetunes/results-{{workflow.parameters.run_name}}/final"
+        when: "{{workflow.parameters.run_inference}} == true"
+
+    - - name: vllm-inference-service
         template: model-inference-service
         arguments:
           parameters:
+            - name: model_uri
+              value: "vllm/{{workflow.parameters.pvc}}/finetunes/results-{{workflow.parameters.run_name}}/final/vllm/model.tensors"
```
dmarx, line 352: flagging to ensure consistency with https://github.com/coreweave/kubernetes-cloud/pull/376/files#r1695728991

dmarx (dismissed):
a couple of minor comments here and there which you can take or leave. LGTM

sangstar (dismissed):
Generally LGTM. How have you tested these changes? Do you have unit tests somewhere or did you do an integration test?

finetuner-workflow/finetune-workflow.yaml
harubaru: After the PR gets merged I'm thinking about making a PR against `ml-containers` to build a finetuner image

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -310,53 +336,90 @@ spec:
...
       securityContext:
         runAsUser: 0
       resources:
         requests:
-          memory: 512Mi
+          memory: 2Gi
           cpu: "2"
         limits:
-          memory: 512Mi
+          memory: 2Gi
```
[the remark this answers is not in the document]
harubaru: The model downloader can run out of memory when downloading large models. I was suggested to use 2Gi instead of 512Mi because of that.

finetuner-workflow/finetune-workflow.yaml
harubaru: I agree! HF changed the env var name from `HF_API_TOKEN` to `HF_TOKEN` at some point. This change would have to go under our `gpt_bpe` repository.

finetuner-workflow/finetuner/finetuner.py
harubaru: That comment is part of a whole sentence, so it does not need to be capitalized. I just checked the file to see if the grammar was off in the comments and it seemed to be fine

finetuner-workflow/finetune-workflow.yaml
harubaru: This was discussed a while ago, the final output path being pretty wonky was a consequence of the serialization script used. However, I think the finetuner should have it's own vLLM serialization script which by the looks of things doesn't seem hard to do.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -582,31 +702,42 @@ spec:
...
+                image: "{{workflow.parameters.inference_image}}:{{workflow.parameters.inference_tag}}"
                 imagePullPolicy: IfNotPresent
-                command: ["/usr/bin/bash", "-c"]
-                args: ["/usr/bin/python3 /app/inference.py --model=${INFERENCE_MODEL} --port=${INFERENCE_PORT}"]
+                command: ["/bin/bash", "-c"]
+                args:
+                  - |
+                    python -m vllm.entrypoints.openai.api_server \
+                      --model ${INFERENCE_MODEL_PATH} \
+                      --model-loader-extra-config "{\"tensorizer_uri\": \"${INFERENCE_MODEL_URI}\"}" \
```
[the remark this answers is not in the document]
harubaru: I would feel more comfortable doing this whenever we update the vLLM serialization image which seems to be related to [this PR](https://github.com/coreweave/ml-containers/pull/72).

finetuner-workflow/finetune-workflow.yaml
harubaru: [Addressed here](https://github.com/coreweave/kubernetes-cloud/pull/376#discussion_r1699151310)

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -184,19 +200,23 @@ spec:
...
     - name: dataset_downloader_tag
       value: 'cd6408a'
     - name: finetuner_image
-      value: 'gooseai/finetuner'
+      value: 'docker.io/harubaru1/finetuner'
     - name: finetuner_tag
-      value: 'b32173f'
+      value: '125'
+    - name: inference_image
+      value: 'docker.io/rtalaricw/vllm'
```
wbrown, line 217: We should create vLLM images hosted on GHA rather than do this.

wbrown (changes requested):
We should be hosting a vLLM container on GHCR, not on Docker.

finetuner-workflow/finetune-workflow.yaml
```diff
@@ -582,31 +702,42 @@ spec:
...
                 ports:
                   - protocol: TCP
                     containerPort: 80
                 livenessProbe:
                   httpGet:
-                    path: /
+                    path: /health
                     port: 80
-                  initialDelaySeconds: 300
+                  initialDelaySeconds: 60
```
Eta0, line 734: Maybe bump these back up a little higher, since it timing out too early would cause more issues than taking a little too long to time out in the error case.

finetuner-workflow/finetuner/Dockerfile
```diff
@@ -1,14 +1,15 @@
...
 # in the final finetuner image.
 # gcc-10/g++-10/lld do not need to be installed here, but they improve the build.
 # gfortran-10 is just for compiler_wrapper.f95.
 FROM ${BASE_IMAGE} as builder
-RUN apt-get install -y --no-install-recommends \
+RUN  apt-get update -y && \
+    apt-get install -y --no-install-recommends \
         cuda-nvcc-11-8 cuda-nvml-dev-11-8 libcurand-dev-11-8 \
         libcublas-dev-11-8 libcusparse-dev-11-8 \
         libcusolver-dev-11-8 cuda-nvprof-11-8 \
```
Eta0, line 15: Probably not needed with our `torch` base images:
- `libcurand`
- `libcublas`
- `libcusparse`
- `libcusolver`

If you are using `torch-extras` then `cuda-nvcc` and `cuda-nvprof` (and most likely `cuda-nvml` unless we use it for other reasons than for building libraries) are probably all also not needed.

finetuner-workflow/finetuner/Dockerfile
Eta0, line 1: Overall, most of this stuff is not necessary when using our `torch-extras` images; try using those as a base and we can probably get rid of most of this Dockerfile and the custom compilation steps in here.

online-inference/hf-llm/service/service.py
Eta0, line 1: These changes aren't relevant to this PR nor needed by it, so can this be reverted / cut out into another branch if we want to keep it?

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -436,7 +443,7 @@ def main_process_print(*args, **kwargs):
     if "eos_token" not in tokenizer.special_tokens_map:
         tokens_to_add["eos_token"] = "<|endoftext|>"
     if "pad_token" not in tokenizer.special_tokens_map:
-        tokens_to_add["pad_token"] = "<|endoftext|>"
+        tokens_to_add["pad_token"] = tokenizer.eos_token
```
Eta0, line 446: This needs to match `gpt_bpe`'s hardcoded default (which should currently be unsigned -1).

Eta0 (changes requested):

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -493,6 +500,26 @@ def compute_loss(self, model, inputs, return_outputs=False):
         return results
 
 
+def tensorizer_save(checkpoint_dir: str) -> None:
+    checkpoint_model = trainer.model
+    supported_classes = (PreTrainedModel,)  # TODO: PEFT support.
+    checkpoint_model = checkpoint_model if isinstance(checkpoint_model, supported_classes) else unwrap_model(
+        checkpoint_model)
```
Eta0, line 507: `unwrap_model` should be able to be called unconditionally here.

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -493,6 +500,26 @@ def compute_loss(self, model, inputs, return_outputs=False):
         return results
 
 
+def tensorizer_save(checkpoint_dir: str) -> None:
+    checkpoint_model = trainer.model
```
Eta0, line 504: The model should probably be an argument instead of pulling it from the `trainer` global.

Eta0 (changes requested):

finetuner-workflow/finetuner/finetuner.py
```diff
@@ -647,14 +674,17 @@ def __init__(self, path: str, context_length: int = 2048):
...
-        self._pad_token_id = tokenizer.pad_token_id
+        self._pad_token_id = 0xffff if use_uint16 else 0xffffffff
+        self._token_dtype = numpy.uint16 if use_uint16 else numpy.uint32
         if is_main_process():
             logger.info(f"DATASET: {path}")
             logger.info(
                 f"DATASET SIZE: {length_mb:,.2f}MiB, {num_tokens:,} tokens, "
                 f"{self.length:,} contexts"
             )
+            logger.info(f"DATASET PAD ID: {self._pad_token_id}")
```
Eta0, line 687: If the `uint32` functionality isn't working here yet, remove it (or comment it out with an explanation) before merging this PR, so it doesn't confuse people.

PR #614, opened 2025-02-04
chore: remove deprecated burstable examples

ChandonPierre:
These are being retired


59psi (approved):
Removing the burstable range so this looks correct as well

